PR #4, opened 2021-03-08
Moving video-producer to content-components

qwangd2l:


rschick (approved):

README.md
```diff
@@ -4,9 +4,10 @@ A collection of components used for content in Brightspace.
 
 ## Components
 
-* [d2l-content-store](components/d2l-content-store): d2l-content-store
-* [d2l-content-uploader](components/d2l-content-uploader): d2l-content-uploader
-* [d2l-capture-central](components/d2l-capture-central): d2l-capture-central
+* [d2l-content-store](components/d2l-content-store)
+* [d2l-content-uploader](components/d2l-content-uploader)
+* [d2l-capture-central](components/d2l-capture-central)
+* [d2l-labs-video-producer](components/d2l-labs-video-producer)
```
rschick, line 10: In a future PR I think we should drop the "labs" to be consistent with the other components in this repo.

AdmiralCoco (approved):

package.json
```diff
@@ -44,6 +44,7 @@
     "@brightspace-ui-labs/pagination": "^1",
     "@brightspace-ui-labs/video-producer": "^0",
     "@chaitin/querystring": "^1.1.0",
+    "@createjs/easeljs": "^2.0.0-beta.4",
```
rschick, line 47: Should these ranges be looser, eg. "^2"? For easeljs and polymer, querystring, etc.

rschick:
In future, would it make sense to put all the langterms together at the top level? Might reduce duplication?

package.json
qwangd2l: For createjs/easljs, I think it needs the beta specification.
As for the other ones, I don't think it matters too much. I'm ok with either

qwangd2l:
> In future, would it make sense to put all the langterms together at the top level? Might reduce duplication?

A lot of the other repos with collection of components have their separate lang terms.
I'm not sure if this is because there's an optimization in place for loading less files or not.

PR #15, opened 2021-03-25
Restructure folders so that we have core and capture subfolders inste…

qwangd2l:
…ad of components.

Add a single line import file for each component in either core/ or capture/

Move d2l-capture-central and d2l-video-produer to capture/
Move d2l-content-store, d2l-content-uploader, d2l-content-service-viewer, d2l-captions-manager to core/


capture/d2l-capture-producer/test/capture-producer.test.js
```diff
@@ -1,19 +1,19 @@
-import '../d2l-content-service-viewer.js';
+import '../d2l-capture-producer.js';
```
rschick, line 1: did something weird happen here?

rschick (approved):

capture/d2l-capture-producer/test/capture-producer.test.js
qwangd2l: Some of the test files are very similar that it thinks the renaming comes from a different file.
For this particular file, there's a paired rename that goes with this incorrect diff:
video-producer/test/video-producer.test.js  -> core/d2l-content-viewer/test/d2l-content-viewer.test.js

PR #22, opened 2021-08-24
US130907 - Add zooming to timeline

Jevan5:
### Problem

The video editor is limited in functionality for large videos, since the precision of time selection was inversely related to the length of the video. Some simple math:

- The width of the timeline is `~1000` pixels
- For a video of length `x` seconds, each pixel would represent `x / 1000` seconds
- For a video that is `1000` seconds or less, each pixel would represent `<= 1` seconds. In this case, any second of the video could be reached by seeking, adding marks, or adding cuts
- **PROBLEM**: For a video longer than `1000` seconds, each pixel would represent `> 1` seconds. In this case, there would be a set of seconds that would be unreachable by seeking, adding marks, or adding cuts

### Solution

To solve this problem, any point along the timeline can be zoomed. This can be done by dragging a slider down from the timeline, which increasingly zooms around the point on the timeline based on the depth to which the slider is dragged down.

The logic for a timeline (other than the UI) is encapsulated in a new file: `timeline.js`. A `Timeline` contains variable `Marks` and `Cuts` along it, represents a static period of time, has a static width, variable point along timeline to zoom around, and a variable zoom multiplier. Ex:
```
const timeline = new Timeline(durationSeconds = 100, widthPixels = 12);
==>
// each pixel is 100 / 12 = ~8 seconds
0s           100s
 |-----------|

timeline.addMarkToTimelineAtPoint(2);
==>
0s           100s
 |-|---------|

timeline.addMarkToTimelineAtPoint(5);
==>
0s           100s
 |-|--|------|

timeline.addCutToTimelineAtPoint(3);
==>
0s           100s
 |-|XX|------|

timeline.pixelsAroundTimelineToZoomAround = 5;
==>
0s           100s
 |-|XX|------|
      ^

timeline.zoomMultiplier = 2;
==>
// each pixel is 50 / 12 = ~4 seconds
20s           70s
  ||XXXX||----
        ^
```

### Demo

https://user-images.githubusercontent.com/24978103/130637771-5bc6c3f5-3328-4405-a3d4-b4f12bf194e9.mov

### TODO

The timeline should stretch and compress vertically so that the zoom slider is always resting directly below the timeline. This will help clarify for users that the timeline is being zoomed.

### DIFF
--- a/capture/d2l-capture-producer/d2l-capture-producer.js
+++ b/capture/d2l-capture-producer/d2l-capture-producer.js
@@ -10,6 +10,8 @@ import constants from './src/constants.js';
 import { InternalLocalizeMixin } from './src/internal-localize-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { selectStyles } from '@brightspace-ui/core/components/inputs/input-select-styles.js';
+import { styleMap } from 'lit-html/directives/style-map';
+import { Timeline } from './src/timeline';
 
 class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 	static get properties() {
@@ -20,6 +22,9 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 			src: { type: String, reflect: true },
 			_loading: { type: Boolean, attribute: false },
 			_videoLoaded: { type: Boolean, attribute: false },
+			_zoomHandleCurrentOffsetX: { type: Boolean, attribute: false },
+			_zoomHandleDepth: { type: Number, attribute: false },
+			_zoomMultiplierDisplayOpacity: { type: Date, attribute: false },
 		};
 	}
 
@@ -42,6 +47,7 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 			}
 
 			.d2l-video-producer-timeline {
+				display: flex;
 				margin-top: 15px;
 			}
 
@@ -58,9 +64,44 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 			}
 
 			#timeline-canvas {
-				border: 1px solid #787878;
+				border: ${constants.CANVAS_BORDER_WIDTH}px solid #787878;
+			}
+
+			#zoom-handle {
+				height: ${constants.ZOOM_HANDLE_HEIGHT}px;
+				pointer-events: none;
+				position: absolute;
+				width: ${constants.ZOOM_HANDLE_WIDTH}px;
+			}
+
+			#zoom-multiplier {
+				pointer-events: none;
+				position: absolute;
+				text-align: center;
+				top: ${constants.CANVAS_CONTAINER_HEIGHT / 2}px;
+				width: 100%;
+			}
+
+			#canvas-container {
+				height: ${constants.CANVAS_CONTAINER_HEIGHT}px;
+				position: relative;
+				width: ${constants.CANVAS_CONTAINER_WIDTH}px;
 			}
 		`];
+	}
+
+	static _clampBetweenMinAndMax(num, min, max) {
+		return Math.max(Math.min(num, max), min);
+	}
+
+	static _getStageXFromPixelsAlongTimeline(pixelsAlongTimeline) {
+		if (pixelsAlongTimeline === null) return null;
+
+		return pixelsAlongTimeline + constants.TIMELINE_OFFSET_X;
+	}
+
+	static _getPixelsAlongTimelineFromStageX(stageX) {
+		return CaptureProducer._clampBetweenMinAndMax(stageX - constants.TIMELINE_OFFSET_X, 0, constants.TIMELINE_WIDTH);
 	}
 
 	constructor() {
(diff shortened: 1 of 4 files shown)


capture/d2l-capture-producer/src/timeline.js
```diff
@@ -0,0 +1,519 @@
...
+	constructor(inSeconds, outSeconds, timeline) {
+		this.in = inSeconds;
+		this.out = outSeconds;
+		this.timeline = timeline;
+		this.displayObject = null;
+	}
+
+	/**
+	 * Determines the pixels along the timeline that this cut begins and ends at.
+	 * @returns {[number, number]} If the cut is on the timeline, returns an array. Otherwise, returns null.
```
Jevan5, line 11: There are many two-dimensional arrays like this in this file. Another common pattern is to return something like this instead, with descriptive names:
```
const returnValue = cut.getPixelsAlongTimeline(); // Can't reference yet, since it could be null
const { inPixels, outPixels } = returnValue;
```

The only reason I chose to return array is that the variable names from this pattern can sometimes conflict with local variable names, like so:
```
const returnValue = cut1.getPixelsAlongTimeline(); // Can't reference yet, since it could be null
const { inPixels, outPixels } = returnValue;

const returnValue2 = cut2.getPixelsAlongTimeline(); // Can't reference yet, since it could be null
const { inPixels, outPixels } = returnValue2; // variable name conflicts
```

Thoughts?

AdmiralCoco: Would it make sense for `getPixelsAlongTimeline` to return an array with null values instead? (e.g., `[null, null]`)

---

>The only reason I chose to return array is that the variable names from this pattern can sometimes conflict with local variable names, like so:

You can rename the destructured variable, e.g.,
```
const { inPixels: cut1InPixels, outPixels: cut2InPixels } = returnValue;
const { inPixels: cut2InPixels, outPixels: cut2InPixels } = returnValue2;
```

Jevan5: Returning the two-element array with 2 null values is a little redundant, since you would only need to check one of the values.

Right, I forgot about that. Then I think returning an object with descriptive names would be better for all these functions.

AdmiralCoco: Sounds good 👍

capture/d2l-capture-producer/src/timeline.js
```diff
@@ -0,0 +1,519 @@
...
+		}
+
+		delete this.timeline._marks[this.seconds];
+
+		return [cutEndingAtMark, cutStartingAtMark];
+	}
+}
+
+class Timeline {
+	constructor(durationSeconds, widthPixels, cuts = [], zoomMultiplier = 1, pixelsAlongTimelineToZoomAround = Math.round(widthPixels / 2)) {
```
AdmiralCoco, line 160: Also consider making the argument an object here?
```suggestion
	constructor({ durationSeconds, widthPixels, cuts = [], zoomMultiplier = 1, pixelsAlongTimelineToZoomAround = Math.round(widthPixels / 2)} = {}) {
```

capture/d2l-capture-producer/src/timeline.js
```diff
@@ -0,0 +1,519 @@
...
+	getTimeBoundsOnTimeline() {
+		const zoomedDuration = Math.round(this.durationSeconds / this.zoomMultiplier);
+
+		const halfZoomedDuration = Math.ceil(zoomedDuration / 2);
+
+		const timeToZoomAround = Math.round(this.durationSeconds * this.pixelsAlongTimelineToZoomAround / this.widthPixels);
+
+		if (timeToZoomAround < halfZoomedDuration) return [0, zoomedDuration];
+		else if (timeToZoomAround + halfZoomedDuration > this.durationSeconds) return [Math.round(this.durationSeconds - zoomedDuration), Math.ceil(this.durationSeconds)];
+		else return [timeToZoomAround - halfZoomedDuration, timeToZoomAround + halfZoomedDuration];
```
AdmiralCoco, line 311: Nit: unnecessary `else`

AdmiralCoco (approved):
Awesome job @Jevan5, looks great!

capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -632,28 +624,61 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 		return seekMode;
 	}
 
+	static _getStageXFromPixelsAlongTimeline(pixelsAlongTimeline) {
+		if (pixelsAlongTimeline === null) return null;
+
+		return pixelsAlongTimeline + constants.TIMELINE_OFFSET_X;
```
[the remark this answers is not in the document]
Jevan5: Need to handle the null case differently than 0, so this should work:
```
return pixelsAlongTimeline === null
  ? null
  : pixelsAlongTimeline + constants.TIMELINE_OFFSET_X;
```

capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -145,153 +194,138 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
...
-	_checkForMarkDelete(event) {
-		if (this._mouseDownStageX && Math.abs(this._mouseDownStageX - event.stageX) === 0) {
-			const markToDelete = this._marks[Math.round(this._currentMark.displayObject.x)];
-			this._stage.removeChild(markToDelete.displayObject);
-			this._currentMark = null;
-
-			delete this._marks[markToDelete.bound];
-			this._recalculateCuts();
-			this._setCursorOrCurrentMark(this._stage.mouseX);
+	static _clampBetweenMinAndMax(num, min, max) {
```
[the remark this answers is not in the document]
Jevan5: I agree with the name change for sure

AdmiralCoco:
When there are a lot of marks close together, what is the behaviour like when interacting with them when zoomed out? e.g.,
image

It is possible for the user to zoom in/out while dragging a mark? And if so, are there any weird interactions that may happen?

Jevan5:
@AdmiralCoco

> When there are a lot of marks close together, what is the behaviour like when interacting with them when zoomed out?

Whichever mark is on top can still be interacted. They can be dragged closer to each other now though, as it is now needed.

> It is possible for the user to zoom in/out while dragging a mark? And if so, are there any weird interactions that may happen?

It should not be possible.

AdmiralCoco (approved):

PR #26, opened 2021-09-03
US131038: Display uploaded SRT and WebVTT files

MarquisLP:
https://user-images.githubusercontent.com/9592685/132411335-edad765b-c715-4a37-a0c5-713d534269b5.mp4

**Captions cue editing controls (edit text, edit timestamps, delete) don't have any functionality yet.**

For optimization and to keep the page from freezing due to DOM overload, I used [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to gradually add captions cues to the DOM whenever the user scrolls down to the bottom of the list. On Firefox, Chrome, and Edge this works well, but on Safari the scrolling begins to lag after the first 100-150 cues. I will try to open a follow-up PR that adds further optimization by hiding cues when they are scrolled out of view.

***

**Previous demo**

https://user-images.githubusercontent.com/9592685/132047834-9af276f0-dd73-4678-9151-9f6970965db4.mov


capture/d2l-capture-producer/scripts/vtt.min.js
```diff
@@ -0,0 +1,3 @@
+/* vtt.js - v0.12.1 (https://github.com/mozilla/vtt.js) built on 03-12-2015 */
```
MarquisLP, line 1: Mozilla's vtt.js library doesn't use ES6 exports, so I had to include the pre-built minified script directly in the project. vtt.js hasn't been updated since 2015 but is still in active use, so it should be safe to hardcode the script like this.

capture/d2l-capture-producer/test/captions-utils.test.js
```diff
@@ -0,0 +1,190 @@
...
+00:06.000 --> 00:08.000
+<v Roger Bingham>from the American Museum of Natural History
+
+00:03.000 --> 00:06.000
+<v Roger Bingham>We’re actually at the Lucern Hotel, just down the street
+
+00:08.000 --> 00:10.000
+<v Roger Bingham>And with me is Neil deGrasse Tyson`;
+
+			const vttParserMock = {};
```
MarquisLP, line 109: vtt.js needs a DOM page when imported, so I mocked it instead to keep these unit tests lightweight.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -1,26 +1,274 @@
...
+			if (propName === 'captionCues') {
+				if (oldValue && oldValue.length === 0 && this.captionCues.length > 0) {
+					const options = {
+						root: this.shadowRoot.querySelector('.d2l-video-producer-captions-cues-list'),
+						rootMargin: '0px',
+						threshold: 0.1,
+					};
+					const listBottom = this.shadowRoot.querySelector('.d2l-video-producer-captions-cues-list-bottom');
+					this.intersectionObserver = new IntersectionObserver(this._updateVisibleCuesInList, options);
+					this.intersectionObserver.observe(listBottom);
```
MarquisLP, line 143: This strategy for infinite scroll was taken from: https://dev.to/trex777/infinite-scrolling-using-intersection-observer-api-118l

The idea is that we have a `<div>` at the bottom of the captions list. When the user scrolls to the bottom of the list, the `<div>` becomes visible, triggering the Intersection Observer to call `_updateVisibleCuesInList()` to add more cue elements to the list.

capture/d2l-capture-producer/src/constants.js
```diff
@@ -19,6 +19,10 @@ const constants = {
 	},
 	CANVAS_WIDTH: 985,
 
+	// Captions
+	MAX_CAPTIONS_UPLOAD_SIZE_IN_BYTES: Math.pow(2, 30), // 1 GB
```
MichaelKovalchuk, line 23: can we make it more obvious? 1024*1024*1024 ?

MichaelKovalchuk:
going to discuss design ideas on Tuesday

MarquisLP:
Added a new demo video to the first comment, showing the new collapsible layout for the captions cues.

MichaelKovalchuk:
looks better now, i would probably make the timer buttons next to the Start and End Text Edits
Also, these text edits, including text area do not look like d2l styled

MichaelKovalchuk (approved):

MarquisLP:
> Also, these text edits, including text area do not look like d2l styled

Thanks for catching that. When I refactored the cue list item into its own LitElement class, I forgot to copy over the D2L styles for textareas. I'll fix that now.

MarquisLP:
Fixed text input styles

Sadly, there isn't enough horizontal space to put the time icons beside the text inputs. I tried shrinking the text inputs and the icon buttons, but the D2L components don't have options for resizing.

MichaelKovalchuk:
Maybe put it next to Start and End?

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -1,26 +1,351 @@
...
+
+	static get styles() {
+		return [ inputStyles, labelStyles, css`
+			.d2l-video-producer-captions-cues-list-item {
+				display: flex;
+				flex-direction: column;
+				background-color: var(--d2l-color-gypsum);
+				border-radius: var(--d2l-input-border-radius, 0.3rem);
+				padding: 12px;
+				margin-bottom: 15px;
```
AdmiralCoco, line 30: Nit: I like to keep styles alphabetically sorted for readability - [VSCode has an extension for this](https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines)

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -1,26 +1,351 @@
...
+	// <script> tags must be added via Javascript in LitElement.
+	// https://stackoverflow.com/a/55693185
+	_appendVttScript() {
+		const script = document.createElement('script');
+		script.src = '../scripts/vtt.min.js';
+		return script;
+	}
+
+	_onFilesAdded(event) {
+		if (event && event.detail && event.detail.files.length === 1) {
```
AdmiralCoco, line 273: Minor: we could invert this statement to return early instead, to reduce nesting and having the code under the if block

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -1,26 +1,351 @@
...
+			const file = event.detail.files[0];
+			const extension = file.name.split('.').pop();
+			const alertToast = this.shadowRoot.querySelector('#d2l-video-producer-captions-alert-toast');
+			if (!['srt', 'vtt'].includes(extension.toLowerCase())) {
+				alertToast.setAttribute('open', 'open');
+				alertToast.setAttribute('type', 'critical');
+				alertToast.innerText = this.localize('captionsInvalidFileType');
```
AdmiralCoco, line 280: Perhaps instead of setting the `alertToast` here, we could add a helper to reuse, e.g.,
```js
_openAlertToast({type, text}) {
    const alertToast = this.shadowRoot.querySelector('...');
    alertToast.setAttribute('...');
}
```

And update it to be:
```suggestion
				this._openAlertToast({type: 'critical', text: this.localize('captionsInvalidFileType')});
```

AdmiralCoco (approved):
Looks great @MarquisLP !

MarquisLP:
> Maybe put it next to Start and End?

Got it to work:

Sync buttons above timestamps

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
MarquisLP: Thanks for suggesting that extension! I'll do that as it would be consistent with how the functions are sorted.

PR #28, opened 2021-09-15
US131039: Implement revision selection in Capture Central Producer

MarquisLP:
This PR adds a revision dropdown that allows the user to load and edit the full list of revisions for a content object. If the user tries to switch revisions while the current revision has unsaved changes, a confirmation dialog is shown.

https://user-images.githubusercontent.com/9592685/133482826-b0af9285-db44-4b3e-9162-0b424fcea4b8.mp4

The UI changes (e.g. changing the save icon button to a "Save Draft" button) were made according to a discussion I had with Michael and Peter.

Right now, Content Service doesn't store timestamps for content revisions. Because of that, the revision dropdown displays the revision indices for now. Once revision timestamps have been added in Content Service, a followup PR will change the dropdown to display pretty-fied timestamps.

Upcoming PRs will:
- Save captions data alongside metadata
- Mirror the UI changes in the [Producer demo page](https://github.com/Brightspace/content-components/blob/master/capture/d2l-capture-producer/demo/demo-video-producer.js)


applications/d2l-capture-central/src/pages/producer/d2l-capture-central-producer.js
```diff
@@ -224,6 +276,32 @@ class D2LCaptureCentralProducer extends DependencyRequester(PageViewElement) {
 		this._selectedLanguage = e.detail.selectedLanguage;
 	}
 
+	async _handleSelectedRevisionChanged(e) {
+		const newlySelectedRevisionIndex = this._revisionsLatestToOldest.length - Number.parseInt(e.target.text.match(/(\d+)/));
```
AdmiralCoco, line 280: Would it be easier if we used [data-attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) on the menu items to indicate the indices?
e.g.,
```js
<d2l-menu-item data-revisionIndex=0>

...

const newlySelectedRevisionIndex = this._revisionsLatestToOldest.length - e.target.dataset.revisionIndex
```
(Happy with your current regex approach, if you think it's added complexity or doesn't improve readability)

AdmiralCoco (approved):
Looks good 👍

applications/d2l-capture-central/src/pages/producer/d2l-capture-central-producer.js
MarquisLP: Wow, thanks for the heads-up on data-attributes. I didn't know about them at all, but I can already see them being very useful. I'll use them here as they'd help reduce the complexity a great deal (not needing regex, not needing to convert from index to 'reverse' index).

PR #29, opened 2021-09-22
US131039: Load/Save captions and metadata internally in Producer

MarquisLP:
This PR:
- Refactors all loading and saving functionality into Producer, rather than handling this logic in Producer's parent (Capture Central in this case).
- Adds the ability to load and save captions for a specific revision and locale.

**For now, captions and metadata are saved to the latest revision as draft files. The draft workflow will be fixed in followup PRs here and in d2l-content-service.**

https://user-images.githubusercontent.com/9592685/134427854-daf33dad-872b-4483-be3e-10e19c8773de.mp4

***

Significant changes were made to captions parsing. Instead of manually parsing a WebVTT file using util functions, we take a URL to the raw WebVTT text and pass it down to the Media Player. The Media Player then passes the text down to a `<track>` element, which automatically parses the text into VTTCue objects. These VTTCues are then passed back up to Producer, which passes it down to Captions Editor.

This has the added benefit of keeping the Captions Editor and the Media Player synchronized: the captions shown in the editor will also appear during media playback.

When an SRT file is uploaded, we convert it to WebVTT text, attach it to a temporary URL using [URL.createObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL), then pass this URL to the Media Player so that it's parsed the same way as a WebVTT file.

***

Currently, captions do not reload when changing the selected language in Producer. This will also be addressed in a future PR.


MarquisLP:
Closing because Sauce Connect for this repo fails on forks. The new PR is #30.

PR #30, opened 2021-09-23
US131039: Load/Save captions and metadata internally in Producer

MarquisLP:
This PR:
- Refactors all loading and saving functionality into Producer, rather than handling this logic in Producer's parent (Capture Central in this case).
    - As a result, Producer now takes in 3 attributes:
        - endpoint
        - tenantId
        - contentId
- Adds the ability to load and save captions for a specific revision and locale.

**For now, captions and metadata are saved to the latest revision as draft files. The draft workflow will be fixed in followup PRs here and in d2l-content-service.**

https://user-images.githubusercontent.com/9592685/134427854-daf33dad-872b-4483-be3e-10e19c8773de.mp4

***

Significant changes were made to captions parsing. Instead of manually parsing a WebVTT file using util functions, we take a URL to the raw WebVTT text and pass it down to the Media Player. The Media Player then passes the URL down down to a `<track>` element, which automatically loads the text and parses it into VTTCue objects. These VTTCues are then passed back up to Producer, which passes it down to Captions Editor.

This has the added benefit of keeping the Captions Editor and the Media Player synchronized: the captions shown in the editor will also appear during media playback.

When an SRT file is uploaded, we convert it to WebVTT text, attach it to a temporary URL using [URL.createObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL), then pass this URL to the Media Player so that it's parsed the same way as a WebVTT file.

***

Currently, captions do not reload when changing the selected language in Producer. This will also be addressed in a future PR.


capture/d2l-capture-producer/demo/demo-video-producer.js
```diff
@@ -93,11 +93,6 @@ class DemoVideoProducer extends LitElement {
 				</div>
 
 				<d2l-capture-producer
-					.defaultLanguage="${this.defaultLanguage}"
-					.metadata="${this.metadata}"
-					.selectedLanguage="${this.selectedLanguage}"
-					@metadata-changed="${this._handleMetadataChanged}"
-					src="http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"
 				></d2l-capture-producer>
```
MarquisLP, line 96: Removed these on the demo page for now so that Lit tests don't break. The demo page will either be revised or removed in another PR.

capture/d2l-capture-producer/scripts/vtt.min.js
```diff
@@ -1,3 +0,0 @@
-/* vtt.js - v0.12.1 (https://github.com/mozilla/vtt.js) built on 03-12-2015 */
```
MarquisLP, line 1: This file is no longer needed because we now parse WebVTT text using the Media Player.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -354,22 +365,42 @@ class VideoProducerCaptions extends InternalLocalizeMixin(LitElement) {
...
+
+	_updateLazyLoadForCaptionsCuesList(oldCaptionsValue) {
+		if (!this.loading && oldCaptionsValue && oldCaptionsValue.length === 0 && this.captions.length > 0) {
+			const options = {
+				root: this.shadowRoot.querySelector('.d2l-video-producer-captions-cues-list'),
+				rootMargin: '0px',
+				threshold: 0.1,
+			};
+			const listBottom = this.shadowRoot.querySelector('.d2l-video-producer-captions-cues-list-bottom');
+			this.intersectionObserver = new IntersectionObserver(this._loadMoreVisibleCuesIfListBottomReached, options);
```
MichaelKovalchuk, line 391: maybe we could call it `listBottomObserver` ? 
then, `_loadMoreVisibleCuesIfListBottomReached` could be renamed to `_loadMoreVisibleCues`

MichaelKovalchuk (approved):

MichaelKovalchuk (approved):

capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -1,29 +1,54 @@
...
-			defaultLanguage: { type: Object },
-			metadata: { type: Object },
-			selectedLanguage: { type: Object },
-			src: { type: String, reflect: true },
+			_alertMessage: { type: String, attribute: false },
+			_captions: { type: Array, attribute: false },
+			_captionsUrl: { type: String, attribute: false },
+			_captionsLoading: { type: Boolean, attribute: false },
+			_content: { type: String, attribute: false },
+			contentId: { type: String, attribute: 'content-id' },
```
AdmiralCoco, line 36: Nit: I prefer keeping the _ properties separate ("private" vs. public properties)

AdmiralCoco (approved):

PR #31, opened 2021-09-23
Add auto captions generation option to upload component.

qwangd2l:
Replace dependency provider/requester with the ones provided from brightspace/core mixins.

The list of languages and their codes depend on the api endpoint `/d2l/api/le/1.58/wcs/autoCaptionsLanguageCodes` added in this LE PR: https://github.com/Brightspace/lms/pull/14369

What it looks like:

https://user-images.githubusercontent.com/37313128/134565081-56ce4638-d99d-4b8e-9c36-cb6a009e572a.mov


MichaelKovalchuk (approved):

MichaelKovalchuk:
can we make dropdown height slightly smaller so the auto-captions checkmark would showup?

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -105,7 +105,8 @@ export class Uploader {
 
 			await this.apiClient.processRevision({
 				contentId: this.content.id,
-				revisionId: this.revision.id
+				revisionId: this.revision.id,
+				captionLanguages,
```
MichaelKovalchuk, line 109: when we create revision, if the language is set, we should also pass sourceLanguage: https://github.com/Brightspace/d2l-content-service/blob/master/api/models/content.js#L49

qwangd2l:
> can we make dropdown height slightly smaller so the auto-captions checkmark would showup?

This is a bit complicated to do since we are using styles provided for select. If we make it smaller, the text doesn't look great either. The amount of size decrease we'd have to make to get the checkmark to show up without scrolling is a lot.

Should we rearrange the UI somehow?

MichaelKovalchuk:
image
So there is no way we can remove the padding here? I marked it red

MichaelKovalchuk:
image
I modded the dialog a bit. To me, it looks just a good a before, but, now we should have the estate for checkmark

qwangd2l:
> image
> So there is no way we can remove the padding here? I marked it red
> All this blank space seems to be unnecessary

Oh, yes we can remove these paddings in the file-drop element. 

Maybe something like this if we just remove all the padding from file-drop element:
Screen Shot 2021-09-23 at 1 57 03 PM

MichaelKovalchuk:
it already looks better to me. just one thing 
image

qwangd2l:
Updated the layout and the demo video in the original PR description.

MichaelKovalchuk (approved):

PR #35, opened 2021-09-29
US131578: Update Producer's revisions UI to match draft workflow 

MarquisLP:
The draft and revisions workflow for Producer X will be as follows:

- **A Producer X session serves as a "canvas" for the latest draft.** On initial page load, Producer X retrieves and displays the draft data for the latest revision. When the user clicks "Save Draft", Producer X will always save to the latest revision's draft.
- **Past revisions can be copied from, but they are never modified.** There is no longer the concept of "switching the currently-loaded revision" using the revisions dropdown. Now, the dropdown simply copies the data from a selected revision onto the current session's "canvas". Saving this copied data will send it to the latest draft, _not_ to the revision it was copied from.
- **When a revision is finished and published, Producer X automatically starts working on the new revision's draft (which carries over the data from the previous revision).**

This PR updates the revisions and loading/saving UI to better communicate this workflow to the user:

- The revisions dropdown has been reworded to make it clear that we are copying past data, not switching between revisions. This includes rewording the warning popup that appears when selecting a revision in the dropdown while the user has unsaved changes.
- A "Saved/Unsaved" indicator has been added at the top of the editor, indicating whether anything has changed compared to the latest draft. One of the things that will trigger the "Unsaved" state is copying from a past revision.

**Captions still do not support drafts; they're currently shared between drafts and non-drafts.** Draft support for captions will be added in followup PRs in this repo and in d2l-content-service.

https://user-images.githubusercontent.com/9592685/135313844-7f6a12f8-0a64-441b-9a6e-3a05b56641d3.mp4


MichaelKovalchuk (approved):

AdmiralCoco:
>If the work is not saved, we should add a Dialog with a warning that user progress may be lost. We can add it in a separate PR, i think

I believe In the current producer, we open in a new window and add a [listener](https://github.com/Brightspace/capture-producer/blob/master/src/capture-producer.js#L63) for `onBeforeUnload`, to confirm before navigating away

Given that this is now within a SPA, the approach we use may be different

capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -1196,6 +1199,38 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
...
+				<p class="d2l-label-text" style="color: var(--d2l-color-feedback-warning);">${this.localize('unsaved')}</p>
+				<d2l-icon class="d2l-video-producer-saved-unsaved-indicator-icon" icon="tier1:alert" style="color: var(--d2l-color-feedback-warning);"></d2l-icon>
+			`;
+		} else {
+			contents = html`
+				<p class="d2l-label-text" style="color: var(--d2l-color-feedback-success);">${this.localize('saved')}</p>
+				<d2l-icon class="d2l-video-producer-saved-unsaved-indicator-icon" icon="tier1:check-circle" style="color: var(--d2l-color-feedback-success);"></d2l-icon>
+			`;
+		}
+		return html`<div class="d2l-video-producer-saved-unsaved-indicator-container">${contents}</div>`;
```
AdmiralCoco, line 1231: Minor: Alternatively, we could move some of the differences to variables I think? - e.g., 

```suggestion
  let icon = 'tier1:check-circle';
  let feedbackColor = 'var(--d2l-color-feedback-success)';
  let langterm = 'saved';
    
  if (this._unsavedChanges) {
    feedbackColor = 'var(--d2l-color-feedback-warning)';
    langterm = 'unsaved';
    icon = 'tier1:alert';
  }
  return html`
    <div class="d2l-video-producer-saved-unsaved-indicator-container">
      <p class="d2l-label-text" style="color: ${feedbackColor};">
        ${this.localize(langterm)}
      </p>
      <d2l-icon class="d2l-video-producer-saved-unsaved-indicator-icon" icon="${icon}" style="color: ${feedbackColor};"></d2l-icon>
    </div>`;
```

I think this reduces the duplication a bit but might be negligible - happy with what you have as well

AdmiralCoco (approved):
Looks great - the copy from previous revision makes a lot of sense as well 👍

MarquisLP:
> Given that this is now within a SPA, the approach we use may be different

That's true. Given that the way to leave Producer X varies depending on where it's is embedded (closing the "Edit Video" dialog in Course Content, clicking Back or on a breadcrumb in Capture Central, etc.), my preliminary idea is to make `unsavedChanged` into a [reflected LitElement property](https://lit-element.polymer-project.org/guide/properties#reflected-attributes) so that the parent can observed it and implement custom confirmation logic.

However, it would be nice to have a solution internal to Producer so that we don't have to duplicate work across different parts of LE. I'll think about this more deeply before opening a PR for it.

PR #40, opened 2021-10-05
Fix lazy load breaking in captions list when copying from revision

MarquisLP:
For lazy loading in the captions list (i.e. displaying more captions cues whenever the user scrolls down to the bottom), we attach an `IntersectionObserver` to a div at the bottom of the list.

However, during a session, this div can be destroyed and recreated. This can happen when copying a past revision, as it replaces the component's entire DOM with a "Loading" indicator. When this happens, the `IntersectionObserver`'s div reference becomes stale, and thus lazy loading stops working.

This PR fixes it by recreating the `IntersectionObserver` whenever the Captions Editor component finishes loading.

### DIFF
--- a/capture/d2l-capture-producer/src/d2l-video-producer-captions.js
+++ b/capture/d2l-capture-producer/src/d2l-video-producer-captions.js
@@ -268,8 +268,16 @@ class VideoProducerCaptions extends InternalLocalizeMixin(LitElement) {
 
 	updated(changedProperties) {
 		changedProperties.forEach((oldValue, propName) => {
-			if (propName === 'captions') {
-				this._updateLazyLoadForCaptionsCuesList(oldValue);
+			if (propName === 'loading' && !this.loading) {
+				// We use IntersectionObserver for lazy loading.
+				// IntersectionObserver uses a direct reference to its root element. In this case, the root
+				// is an empty div we use to mark the bottom of the captions cues list.
+				// Whenever this empty div is recreated (e.g. after the component goes from loading to non-loading state),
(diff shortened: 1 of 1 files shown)


capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -391,16 +402,17 @@ class VideoProducerCaptions extends InternalLocalizeMixin(LitElement) {
...
-			this._numberOfVisibleCuesInList = constants.NUM_OF_VISIBLE_CAPTIONS_CUES;
-		} else if (this.captions.length < this._numberOfVisibleCuesInList) {
+		if (
+			(Math.abs(this.captions.length - this._numberOfVisibleCuesInList) === 1) || // Adding or removing a single captions cue through the UI
+			(this.captions.length <= constants.NUM_OF_VISIBLE_CAPTIONS_CUES)
+		) {
 			this._numberOfVisibleCuesInList = this.captions.length;
+		} else if (this.captions.length > constants.NUM_OF_VISIBLE_CAPTIONS_CUES) {
+			this._numberOfVisibleCuesInList = constants.NUM_OF_VISIBLE_CAPTIONS_CUES;
 		}
```
MarquisLP, line 416: The old logic would only update the number of visible cues when the length of the captions array goes from zero to non-zero (i.e. after uploading captions). But now that the user can copy from a past revision at any time, it is possible for the captions array to change from a non-zero length to another non-zero length.

MichaelKovalchuk (approved):

AdmiralCoco (approved):

PR #44, opened 2021-10-07
Uploader creates content with sharedWith attribute

ktonon:
Content uploaded this way is missing org unit context.

upload-shared-with


ktonon:
I'll submit a separate PR to the lms to provide the value to `<d2l-content-uploader ...>`

PR #52, opened 2021-10-20
Producer: Disable and hide chapters + timeline for non-Video content

MarquisLP:
If the content object is non-video, metadata API calls will be skipped when loading and saving data.

Captions Editor X - Audio Screenshot


MarquisLP:
> Styling not a tab to look like a tab is too complex. How does it work if you always have Chapters tab in html, but it is hidden?

I tried `visibility: hidden` and `display: gone`. In both cases, the "Table of Contents" tab button still appears at the top but its content is empty:

Screen Shot 2021-10-20 at 1 20 31 PM

So it doesn't look like there's an easy way to manipulate the tab buttons.

MichaelKovalchuk (approved):

PR #80, opened 2021-11-03
fix: Cosmetic fixes in Producer X

MarquisLP:
- Fixed a bug where Producer's loading overlay was wider than Producer itself, causing the loading spinner to appear on the right edge of the page instead of the center.
- Fixed a bug where the timeline would appear before Producer finished loading.
- In the captions cue list item layout, the "Start" and "End" labels are now horizontally aligned with the edge of the text input box.
- To prevent confusing the user with two progress bars (Producer's timeline and Media Player's seek bar), Producer passes a `hide-seek-bar` attribute to the Media Player. This attribute will be implemented in a followup PR in the Media Player repo: https://github.com/BrightspaceUILabs/media-player/pull/114

### DIFF
--- a/capture/d2l-capture-producer/d2l-capture-producer-editor.js
+++ b/capture/d2l-capture-producer/d2l-capture-producer-editor.js
@@ -36,6 +36,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement)) 
 			metadataLoading: { type: Boolean, attribute: 'metadata-loading' },
 			selectedLanguage: { type: Object },
 			src: { type: String },
+			timelineVisible: { type: Boolean, attribute: 'timeline-visible' },
 
 			_activeCue: { type: Object, attribute: false },
 			_zoomMultiplier: { type: Number, attribute: false },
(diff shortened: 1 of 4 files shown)


capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -181,6 +183,8 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 						controls
 						crossorigin="anonymous"
 						@cuechange="${this._handleCueChange}"
+						hideseekbar
+						hide-seek-bar
```
AdmiralCoco, line 187: Is this meant to be repeated twice?

AdmiralCoco (approved):

MichaelKovalchuk (approved):

capture/d2l-capture-producer/d2l-capture-producer-editor.js
MarquisLP: Oh, thanks for catching that. I meant to change the initial name (hideseekbar) to the finalized name (hide-seek-bar).

ktonon (approved):

PR #86, opened 2021-11-05
feat: Allow users to switch revisions in Producer while current revision is processing

MarquisLP:
- The loading indicator and "revision is processing" message are no longer overlays. This fixes an issue where the loading/processing overlay would block all page interaction.
- During processing, the revisions dropdown is still enabled. This allows users to switch between revisions while processing and save new drafts.
- In the revisions dropdown, revisions that are processing (i.e. have `ready: false`) are marked with "(processing)".
- If the user stays on the page while the selected revision is processing, and progress polling determines that processing failed, a "processing failed" message is displayed.
- When Producer loads the content object and its revisions from Content Service, it will automatically filter out revisions that have `processingFailed: true`.

`processed-video` was broken on my local dev environment, so unfortunately the demo below doesn't show a processing revision becoming ready. However, I verified that it works by manually editing DynamoDb while Producer was polling for ProgressInfo.

https://user-images.githubusercontent.com/9592685/140586699-a3ba636a-a472-4009-b986-7f9fd89fb252.mp4

Below is the "processing failed" error message:

Producer X - Processing Failed Message


AdmiralCoco (approved):

MichaelKovalchuk (approved):
nice!

PR #87, opened 2021-11-08
Work around for missing statusText in response

ktonon:
I'm using this in local dev, because for some reason the response `statusText` value is an empty string.

That would break my local producer, causing an error message **An error occurred while loading chapters and timeline cuts**, preventing the timeline from loading, and breaking the save draft and finish buttons

### DIFF
--- a/capture/d2l-capture-producer/d2l-capture-producer.js
+++ b/capture/d2l-capture-producer/d2l-capture-producer.js
@@ -602,7 +602,7 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 			});
 			this._captionsUrl = res.value;
 		} catch (error) {
-			if (error.message === 'Not Found') {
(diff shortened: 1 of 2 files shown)


ktonon:
Don't know if we want to merge this, or to figure out why the responseText is empty

MarquisLP (approved):
I'm okay with merging this, as checking the status code is probably more reliable. I've tested these changes on my end and they work fine.

That said, we might still want to investigate the empty `statusText` in case that points to an issue in Content Service. I'll leave that to @MichaelKovalchuk to decide though.

MichaelKovalchuk (approved):

MichaelKovalchuk:
@ktonon I would merge this fix for now and create a trello card to investigate the empty `statusTests`

ktonon:
Tracking with https://trello.com/c/VVztBMDL/130-content-service-empty-statustext-in-response

PR #92, opened 2021-11-10
fix: Fix race condition causing captions to get stuck loading

MarquisLP:
Before, we would start fetching the source audio/video file and the captions file at the same time.

However, the captions need to be parsed by the Media Player. Because of that, if the captions are fetched before the Media Player finishes loading, the captions never get parsed and captions loading never finishes.

The solution is to wait for the Media Player's "loadeddata" event before fetching captions.


capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -296,6 +296,10 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 		});
 	}
 
+	get _editor() {
+		return this.shadowRoot.querySelector('d2l-capture-producer-editor') || null;
```
AdmiralCoco, line 300: Nit: I think `querySelector` returns `null` [if there are no matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector#return_value)

AdmiralCoco (approved):

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -270,6 +270,10 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 		}
 	}
 
+	get mediaPlayer() {
+		return this._mediaPlayer || null;
```
MichaelKovalchuk, line 274: same here? this._mediaPlayer is null if not present

MarquisLP: Whoops, merged already 😅 Thanks for catching that; I'll see if I can include this cleanup in my next fix PR.

PR #93, opened 2021-11-10
fix: advanced editing dialog

MichaelKovalchuk:
image

### DIFF
--- a/core/d2l-content-uploader/locales/en.js
+++ b/core/d2l-content-uploader/locales/en.js
@@ -1,5 +1,6 @@
 export const val = {
 	addSelectedContentToCourse: 'Add your selected content to the course',
+	advancedEditing: 'Advanced Editing',
 	audioLanguageLabel: 'Audio Language:',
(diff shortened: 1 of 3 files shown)


MarquisLP:
Is the dialog scrollable? The timeline doesn't seem to be visible in the screenshot.

MichaelKovalchuk:
> Is the dialog scrollable? The timeline doesn't seem to be visible in the screenshot.

I am going to look into this dialog after. It is the MVC component that we use, and, by default, iframe has `scrollable=no` property. I need to either add a change in LE to make it work, or pass in some property. Planning to do that in the follow up PRs

MarquisLP (approved):

MichaelKovalchuk:
> > Is the dialog scrollable? The timeline doesn't seem to be visible in the screenshot.
> 
> I am going to look into this dialog after. It is the MVC component that we use, and, by default, iframe has `scrollable=no` property. I need to either add a change in LE to make it work, or pass in some property. Planning to do that in the follow up PRs

fix: https://github.com/Brightspace/content-components/pull/98

PR #95, opened 2021-11-11
fix: Redefine Cuts so that 'out: undefined' means end of the media file, to prevent processing errors on end-of-video cuts

MarquisLP:
Producer's timeline works with integral seconds (for conversion to pixels). This means that Cuts can have in and out timestamps of '3' and '5', but not '3.2' or '5.11'.

This can cause file processing issues when there is a cut that finishes at the end of the video. For example, if that Cut's 'out' time is 10, but the media file is actually only 9.95 seconds long, the processing workflow will fail because it will try to work with a timecode (10) that doesn't exist in the media file.

The solution is to remove the 'out' parameter for any Cut that finishes at the end of the media file. (e.g. `{ in: 7}` would mean "cut from 7 seconds to the end".) Our audio/video processing workflows already understand 'out: undefined' to mean 'end of the file'.

More details can be found in this Trello card: https://trello.com/c/Ztzzj225


MarquisLP:
Some timeline tests need to be updated; will push that soon.

AdmiralCoco (approved):

capture/d2l-capture-producer/test/timeline.test.js
```diff
@@ -29,7 +29,7 @@ describe('timeline.js', () => {
 						const cut = t.addCutAtPoint(pixelsAlongTimeline);
 
 						assert.equal(cut.in, 0);
-						assert.equal(cut.out, durationSeconds);
+						assert.isUndefined(cut.out);
```
MichaelKovalchuk, line 32: maybe we should check if we have a property `out`?

MichaelKovalchuk: i.e. `expect(cut).to.not.have.property('out')`

MichaelKovalchuk (approved):

PR #96, opened 2021-11-11
fix: Prevent metadata loading forever when chapters+cuts are disabled

MarquisLP:
The "Save Draft" and "Finish" buttons are disabled while metadata is loading, so this issue caused those buttons to never become enabled for Audio files.


capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -662,6 +662,7 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 		}, { once: true });
 
 		if (this._enableCutsAndChapters) this._loadMetadata(this._selectedRevision.id);
```
MichaelKovalchuk, line 664: nit: maybe a better styling like this: 
```
if (this._enableCutsAndChapters) {
   this._loadMetadata(this._selectedRevision.id);
} else {
   this._metadataLoading = false;
}
```

MichaelKovalchuk (approved):

PR #99, opened 2021-11-15
Metadata support with hypermedia APIs

muhammadzaidi:


core/d2l-content-viewer/src/clients/hypermedia-client.js
```diff
@@ -42,6 +42,21 @@ export default class HypermediaClient {
...
+		if (!resourceEntity.hasActionByName('get-metadata')) {
+			return null;
+		}
+
+		try {
+			const getMetadataAction = resourceEntity.getActionByName('get-metadata');
+			const metadataResponse = await this._fetch({ url: getMetadataAction.href });
+			const metadataEntity = SirenParse(metadataResponse);
+			return metadataEntity.properties;
+		} catch (e) {
```
MichaelKovalchuk, line 55: why do we need to have a `try/catch`  block here?

muhammadzaidi: I added that to avoid a warning about uncaught error thrown by `_fetch` when there is a 404

MichaelKovalchuk: is there a way to check for `Cause` or error code 404 here and throw an error if not?

muhammadzaidi: Could check for `e.message === "Not Found"` then if it's false `throw new Error(e.message);`

MichaelKovalchuk: do you have an error code somewhere in the error?

muhammadzaidi: Doesn't seem like the error has the code. From [here](https://github.com/Brightspace/content-components/blob/6a09a017010ceb4f1ffaf99ee376e95f1e2cff9c/core/d2l-content-viewer/src/clients/hypermedia-client.js#L111) it would just have the `response.statusText`, for 404 that would be `Not Found`.

MichaelKovalchuk: ouch. `Not Found` it is then

core/d2l-content-viewer/src/clients/hypermedia-client.js
```diff
@@ -53,6 +53,7 @@ export default class HypermediaClient {
 			const metadataEntity = SirenParse(metadataResponse);
 			return metadataEntity.properties;
 		} catch (e) {
+			if (e.message !== 'Not Found') throw new Error(e.message);
```
MichaelKovalchuk, line 56: maybe
```
if (e.message !== 'Not Found') throw e;
```
?

MichaelKovalchuk (approved):

rcausey (approved):

PR #105, opened 2021-11-18
US133334: Update Producer to handle changes to unadjusted captions API

MarquisLP:
- Falling back to adjusted captions is now handled on the Content Service backend. So this PR reverts #103.
- The DELETE captions route no longer takes an "adjusted" parameter, so Producer's calls to that route have been modified accordingly.

The Content Service changes can be found in https://github.com/Brightspace/d2l-content-service/pull/1745.

### DIFF
--- a/capture/d2l-capture-producer/d2l-capture-producer.js
+++ b/capture/d2l-capture-producer/d2l-capture-producer.js
@@ -555,6 +555,7 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 				revisionId: revision.id,
 				locale,
 				draft: true,
+				adjusted: false,
 			});
 			this._captionsUrl = res.value;
 		} catch (error) {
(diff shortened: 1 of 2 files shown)


capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -556,7 +555,6 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 				revisionId: revision.id,
 				locale,
 				draft: true,
-				adjusted: !revisionCaptionLocale.unadjustedHash,
```
MichaelKovalchuk, line 559: i think it should still check the same as before. E.g. you should be able to get `adjusted: false` all the time

MichaelKovalchuk: removing this line will make it work identical to the player - the version you receive will always be adjusted

MarquisLP: [apiClient.getCaptionsUrl() sets "adjusted=false" by default](https://github.com/Brightspace/content-components/blob/master/capture/d2l-capture-producer/src/content-service-client.js#L53). Though I suppose that might look confusing.

MichaelKovalchuk: we need to make it `adjusted = true` in the client to be consistent with the Content Service API

MichaelKovalchuk (approved):

PR #108, opened 2021-11-19
fix: Fix Media Player load error caused by passing empty src attribute

MarquisLP:
When Producer X loads, there is a brief period of time where the "src" attribute it passes to the Media Player is an empty string. This is because it needs to load the signed URL from the backend.

However, the Media Player's "src" attribute should never be empty (unless `<source>` tags are passed in).  
In the past, this only resulted in a console warning message, but recent changes to Media Player's handling of sources have made this requirement more strict. (see https://github.com/BrightspaceUILabs/media-player/pull/115)

To address this, we do not render the editor (which contains the Media Player) until we've finished loading the src string.  
While loading the src string, we display the loading indicator instead of the editor.

This commit also prevents a race condition caused by attempting to load captions before the editor has finished its initial render. (Captions are parsed by the Media Player, so we need to wait for the editor and its Media Player to finish loading before fetching captions.)


capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -49,7 +49,6 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 			_src: { type: String, attribute: false },
 			_timelineVisible: { type: Boolean, attribute: false },
 			_unsavedChanges: { type: String, attribute: false },
-			_mediaLoaded: { type: Boolean, attribute: false },
```
MarquisLP, line 52: Removed this property because it isn't being used anywhere anymore.

rcausey:
Hey Mark, I also ran into this issue today. I like this fix, but I was also thinking of making some changes to media-player so it doesn't break when an empty `src` is passed in. 

I think the issue in media-player is with the type detection code - it basically determines whether to render the video or audio player by trying to load the source in the video player first, checking whether the `videoHeight` property is zero, and re-rendering with the audio player if it is. When it gets an empty source, it gets into a bad state, and things break. This type of detection seems really hacky to me, so I want to add a `mediaType` attribute to the media-player element that can be used to specify whether to render the video or audio player and completely get rid of the detection code. I will need to change instances where we render the player so the media type is passed in, but I'm thinking it's worth it. Does this seem reasonable to you? I can't think of any scenarios where we won't know the type when rendering the media player.

MarquisLP:
Hey Rhys, I think that definitely sounds reasonable! I agree that the current type detection is very hacky, and adding a `mediaType` attribute sounds like the best way to improve it.

> I can't think of any scenarios where we won't know the type when rendering the media player.

Yeah, for Content Service objects, we'll always have the type available in the revision object. So for Producer X and other Content Service components, adding `mediaType` should be straightforward. Not sure about local file system audio/videos, but I'm sure it's possible too.

***

So if we go ahead with adding the `mediaType` attribute, what will be the expected behaviour when an empty `src` is passed in?

rcausey:
Ok sounds good, I'll go ahead with that.

I'll make it so that a loading spinner is rendered if no `src` is passed in (or `mediaType` isn't yet set).

rcausey (approved):

MarquisLP:
> I'll make it so that a loading spinner is rendered if no src is passed in (or mediaType isn't yet set).

Gotcha. I'll add `mediaType` and revert back to letting empty src get passed down to the Media Player. I'll hold off on merging this PR until we can test this with the new type detection.

MarquisLP:
Closing this PR because the fix is included in #109.

PR #110, opened 2021-11-22
Remove unused dependency on video-producer

rcausey:
Will also bump minor version


rcausey:
video-producer dependency no longer used and depends on v1.x of media-player

MichaelKovalchuk (approved):

MarquisLP (approved):

PR #130, opened 2021-12-01
restart upload on network error and fix localize issue

gordon-plumb:
- retrying the upload to s3 when a network error occurs
- added `InternalLocalizeMixin` to fix issue where uploader hangs on the progress view after an error, should now go back to upload view with the localized error message

### DIFF
--- a/core/d2l-content-uploader/src/util/s3-uploader.js
+++ b/core/d2l-content-uploader/src/util/s3-uploader.js
@@ -29,7 +29,12 @@ export class S3Uploader {
 	async upload() {
 		const { file, key } = this;
 		const signResult = await this.signRequest({ file, key });
-		return new Promise((resolve, reject) => {
+		let resolve, reject;
+		const promise = new Promise((res, rej) => {
+			resolve = res;
+			reject = rej;
+		});
+		const run = async() => {
 			const xhr = this.createRequest('PUT', signResult.signedUrl);
 			xhr.addEventListener('load', () => {
 				if (xhr.status >= 200 && xhr.status <= 299) {
@@ -39,7 +44,9 @@ export class S3Uploader {
 				}
 			});
 			xhr.addEventListener('error', () => {
-				reject(new Error(`XHR error for ${file.name}: ${xhr.status} ${xhr.statusText}`));
+				setTimeout(() => {
+					run();
+				}, 5000);
 			});
 			xhr.upload.addEventListener('progress', event => {
 				if (event.lengthComputable) {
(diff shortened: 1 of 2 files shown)


core/d2l-content-uploader/src/util/s3-uploader.js
```diff
@@ -39,7 +44,9 @@ export class S3Uploader {
 				}
 			});
 			xhr.addEventListener('error', () => {
-				reject(new Error(`XHR error for ${file.name}: ${xhr.status} ${xhr.statusText}`));
+				setTimeout(() => {
```
gordon-plumb, line 47: Would it be worth adding a limit to the number of retries and displaying an error after the limit is reached?

core/d2l-content-uploader/src/util/s3-uploader.js
```diff
@@ -29,7 +29,12 @@ export class S3Uploader {
 	async upload() {
 		const { file, key } = this;
 		const signResult = await this.signRequest({ file, key });
-		return new Promise((resolve, reject) => {
+		let resolve, reject;
+		const promise = new Promise((res, rej) => {
+			resolve = res;
+			reject = rej;
+		});
+		const run = async() => {
```
MichaelKovalchuk, line 37: i do not think you should use async here

MichaelKovalchuk: there is no await in the function and it does not return promise

core/d2l-content-uploader/src/util/s3-uploader.js
MarquisLP: I'd say yes. Otherwise, if there's a real error going on in the backend, the user may not know when to stop waiting.

core/d2l-content-uploader/src/util/s3-uploader.js
```diff
@@ -67,19 +79,19 @@ export class S3Uploader {
...
+	async _uploadWithRetries(file, signResult, retries = RETRIES) {
+		try {
+			return await this._startUpload(file, signResult);
+		} catch (error) {
+			if (error.name !== 'XHRError' || retries <= 0) {
+				throw error;
+			}
+		}
+		await sleep(randomizeDelay(5000, 1000));
+		return this._uploadWithRetries(file, signResult, --retries);
```
MichaelKovalchuk, line 94: it is a little unclear why we are having retry outside of the try/catch. Would it be nicer to have:
```
try {
			return await this._startUpload(file, signResult);
		} catch (error) {
			if (error.name !== 'XHRError' || retries <= 0) {
				throw error;
			}
           	await sleep(randomizeDelay(5000, 1000));
		    return this._uploadWithRetries(file, signResult, --retries);
		}
```

MichaelKovalchuk (approved):

MarquisLP (approved):

PR #136, opened 2021-12-07
Fix dialog size

MichaelKovalchuk:
Max dialog size is set to 1200px
If the window was resized - reset dialog location
min-width: 170px is required to fix broken button styles in Firefox

### DIFF
--- a/core/d2l-content-uploader/src/components/preview.js
+++ b/core/d2l-content-uploader/src/components/preview.js
@@ -119,20 +119,32 @@ export class Preview extends MobxReactionUpdate(RequesterMixin(InternalLocalizeM
 
 	async _onAdvancedEditing() {
 		const location = `/d2l/le/contentservice/producer/${this._contentId}/view`;
+		let dialog = null;
+		let topWindow = null;
+
+		const onResize = () => {
+			dialog.style.width = 'calc(100% - 50px)';
+			dialog.style.maxWidth = '1200px';
+			dialog.style.left = `${(topWindow.innerWidth - dialog.offsetWidth) / 2}px`;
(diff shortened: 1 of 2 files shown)


MichaelKovalchuk:
image

core/d2l-content-uploader/src/components/preview.js
```diff
@@ -119,20 +119,32 @@ export class Preview extends MobxReactionUpdate(RequesterMixin(InternalLocalizeM
...
 				Onload: (handle) => {
 					setTimeout(() => {
-						handle.dialog.style.width = 'calc(100% - 50px)';
-						handle.dialog.style.left = '25px';
-						handle.dialog.style.top = `calc(${handle.win.top.scrollY}px + 25px)`;
-						handle.dialog.style.height = 'calc(100% - 50px)';
+						dialog = handle.dialog;
+						topWindow = handle.win.top;
+						onResize();
+						topWindow.addEventListener('resize', onResize);
```
AdmiralCoco, line 142: Alternatively (to avoid adding the two extra variables) we could change `onResize` to accept the dialog/topWindow as arguments
```js
onResize(handle.dialog, handle.win.top);
topWindow.addEventListener('resize', onResize.bind(this, handle.dialog, handle.win.top));
```

AdmiralCoco (approved):

core/d2l-content-uploader/src/components/preview.js
MichaelKovalchuk: i tried that, but i still need topWindow variable to remove the listener

AdmiralCoco: Oops missed that - makes sense then

MarquisLP (approved):

PR #144, opened 2021-12-08
Use extensions instead of CS supported mime types to determine which video/audio files can be uploaded

rcausey:
I found that you can use extensions instead of mime types, so I think that's a better option since mime types are inconsistent between browsers. I tested in Chrome, Safari, and Firefox in local dev.


qwangd2l (approved):

MarquisLP (approved):
In the long term, I wonder if it would be a good idea to introduce a new API route for supported file extensions. We could give it a query parameter for specifying audio, video, etc.

rcausey:
> In the long term, I wonder if it would be a good idea to introduce a new API route for supported file extensions. We could give it a query parameter for specifying audio, video, etc.

Yeah I think that's a good idea. I want to minimize changes for now since we're so close to release, but Michael and I will think of that during hygiene planning

AdmiralCoco (approved):

PR #145, opened 2021-12-09
File drop doesn't support 'accept' attribute with extensions

rcausey:


rcausey:
Other file types still get rejected via onFileDrop handler.

MichaelKovalchuk (approved):

qwangd2l (approved):

PR #147, opened 2021-12-09
Use release listener instead

MichaelKovalchuk:


AdmiralCoco (approved):
Did a quick test on my local as well, this does fire on close of the advanced editing dialog

PR #148, opened 2021-12-10
Mkovalchuk/validate srt vtt on upload

MichaelKovalchuk:


capture/d2l-capture-producer/src/constants.js
```diff
@@ -21,7 +21,7 @@ const constants = {
 
 	// Captions
 	ADD_NEW_CUE_KEY_CODE: 13, // 13 is the "Enter" key in Javascript
-	MAX_CAPTIONS_UPLOAD_SIZE_IN_BYTES: 1024 * 1024 * 1024,
+	MAX_CAPTIONS_UPLOAD_SIZE_IN_BYTES: 15 * 1024 * 1024, // 15MB
```
MichaelKovalchuk, line 24: the absolute maximum the video tag can handle is 4000 characters by 5000 cues which, in most default encodings we use 8-16 bytes per character. The maximum VTT file size video tag supports is ~153MB. However, I am making a call that people will never upload VTT files larger that 15MB - 5000 cues 400 characters each if we use 16 bit per character.

rcausey (approved):

PR #153, opened 2021-12-10
Standardize how we determine audio/video type in uploader code

rcausey:


rcausey:
This was causing .ogg failure because it was determining in some areas of code that it wasn't an audio file

MichaelKovalchuk (approved):

core/d2l-content-uploader/src/util/media-type-util.js
```diff
@@ -0,0 +1,18 @@
+const SUPPORTED_AUDIO_EXTENSIONS = ['.m4a', '.mp3', '.ogg', '.wav', '.wma'];
+const SUPPORTED_VIDEO_EXTENSIONS = ['.avi', '.f4v', '.flv', '.m4v', '.mov', '.mp4', '.webm', '.wmv'];
+
+export function getExtension(filePath) {
```
MichaelKovalchuk, line 4: minor: we use file.name or fileName everywhere we access these methods, maybe we should use fileName instead of filePath here?

rcausey: I was thinking filePath here because it can work for file names or paths.

PR #156, opened 2021-12-12
Timeline is longer than duration of the video

MichaelKovalchuk:


MarquisLP (approved):

MarquisLP:
Tested this out locally using the demo, and it appears to work well without the rounding 👍

PR #157, opened 2021-12-13
Producer: ask before exit if there are unsaved changes

ktonon:
Want to keep the changes minimal. This is to support https://trello.com/c/G6xrWO0F/273-producer-x-you-can-close-the-producer-and-lose-your-changes-without-getting-any-warning (Peter's request)

When there are no changes, the user can close the tab or leave the page __without__ a prompt.
Screen Shot 2021-12-13 at 9 08 01 AM

When there are changes this browser native dialog will appear:
Screen Shot 2021-12-13 at 9 09 09 AM

Clicking "Cancel" will do nothing, and the user can manually save. Clicking "Leave" will force the page to unload and unsaved changes are lost.

Peter wanted the dialog to auto save, but that would have been more complicated and possibly introduce bugs.

### DIFF
--- a/capture/d2l-capture-producer/d2l-capture-producer.js
+++ b/capture/d2l-capture-producer/d2l-capture-producer.js
@@ -169,6 +169,8 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 			await this._setupLanguages();
 			await this._loadContentAndAllRelatedData();
 		});
+
+		window.addEventListener('beforeunload', this._askBeforeExit.bind(this));
 	}
 
 	render() {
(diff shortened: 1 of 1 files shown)


ktonon:
Only tested in chrome so far. I'll run through each browser before merging

AdmiralCoco (approved):

MarquisLP (approved):

ktonon:
Testing that the prompt appears when:
- browser back button
- closing tab
- navigating to new page

macOS:
- Chrome: ok
- Firefox: ok
- Safari: only appears when closing the tab, otherwise the user can leave the page without a prompt

MichaelKovalchuk:
I am not sure if it would work in Content Tool and Lessons Producer dialogs

ktonon:
> I am not sure if it would work in Content Tool and Lessons Producer dialogs

Good point. I'll test that too

AdmiralCoco:
I can try and test on local lessons

ktonon:
Also tested Edge on Windows. It was ok for the full page mode

ktonon:
Yeah, @MichaelKovalchuk doesn't work to prevent the dialog closing

AdmiralCoco:
Works for me on local Lessons (minus the dialog)

ktonon:
Going to merge this for now. We can look into handling edge cases like "dialog close" in the next release, as the fix will be too large of a change for this late in the release

PR #164, opened 2021-12-14
Fixed bug with textarea storing the last input value

MichaelKovalchuk:


capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -585,17 +585,18 @@ class VideoProducerCaptions extends InternalLocalizeMixin(LitElement) {
...
-				${this.captions.slice(cuesListStart, cuesListEnd).map(cue => html`
-					<d2l-video-producer-captions-cues-list-item
-						?active="${this.activeCue && (cue.startTime === this.activeCue.startTime) && (cue.endTime === this.activeCue.endTime)}"
-						.cue="${cue}"
-					></d2l-video-producer-captions-cues-list-item>
-				`)}
-				${((this._currentVisibleCueBatch + 1) * constants.NUM_OF_VISIBLE_CUES) < this.captions?.length ? this._renderNextCuesBatchButton() : ''}
-			</div>
-			${this._renderingCues ? this._renderPreparingCaptionsOverlay() : ''}
+			${this._renderingCues ? this._renderPreparingCaptionsOverlay() :
```
MarquisLP, line 588: Oh, I see -- the old cues page wasn't getting torn down when loading the next page, so any cue text input that has focus won't get updated. Explicitly destroying and recreating the cues list fixes that. Nice 👍

MichaelKovalchuk: Yeah, the cues were all up-to-date with the valid data there. the text-input, on the other hand, still stored the value left from the previous page text edit. I am going to remove `id` from the cues, as I do not think those are necessary. I though by adding them - LIT will know that the elements are new, but it did not care :)

MarquisLP: > I though by adding them - LIT will know that the elements are new, but it did not care :)

Haha yeah, I think when I added the id's, I was thinking along the lines of React's `key` attributes. That would be a nice thing for Lit to support.

PR #169, opened 2021-12-15
fix: Show message & top bar when Producer opens on a processing revision

MarquisLP:
This fixes the issue where launching Producer on a processing revision shows only a loading spinner with no 'processing' message or any other controls.  
It also fixes a minor bug where loading another revision would cause the top bar to briefly disappear.

The cause was that we were using a single 'loading' variable for two separate statuses:

- The content object is not yet loaded (only happens at startup)
- The selected revision is not yet loaded

By creating two separate 'loading' variables, we can allow the 'processing' message and the top bar to be shown after the content has loaded, but before the selected revision has finished processing and loading.


capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -176,16 +176,18 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 	}
 
 	render() {
-		const isLoading = !this._selectedRevision || !this._src;
-		const isProcessing = !isLoading && !this._selectedRevision.draft && !this._selectedRevision.ready;
-		const isProcessingFailed = !isLoading && this._selectedRevision.processingFailed;
-		const isReadyForEdit = [isLoading, isProcessing, isProcessingFailed].every(statusFlag => !statusFlag);
+		const isLoadingContent = !this._content || !this._revisionsLatestToOldest;
```
MichaelKovalchuk, line 179: if we get revisionLatestToOldest we get from the content? can we simply put: 
`const isLoadingRevisions = !this._revisionsLatestToOldest` ?

capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -176,16 +176,18 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
...
+		const isProcessing = !isLoadingContent && !this._selectedRevision.draft && !this._selectedRevision.ready;
```
MichaelKovalchuk, line 180: we are not using _selectedRevisionIsProcessing

MichaelKovalchuk:
hi Mark, I used your PR as a platform for fixing Firefox preview bug here: https://github.com/Brightspace/content-components/pull/170

MarquisLP:
> hi Mark, I used your PR as a platform for fixing Firefox preview bug here: #170

Cool, thanks for letting me know. I'll close this PR.

PR #188, opened 2022-01-14
Move methods for reusability

clemencetbk:
Move some methods out of ContentList and into a new class so that RecycleBin and ContentList can share those methods (will add RecycleBin in a separate PR)

### DIFF
--- a/applications/d2l-capture-central/src/components/capture-central-list.js
+++ b/applications/d2l-capture-central/src/components/capture-central-list.js
@@ -156,8 +156,8 @@ export class CaptureCentralList extends DependencyRequester(InternalLocalizeMixi
 		this.loading = true;
 		const { sortQuery, searchQuery, dateModified, dateCreated } = this.queryParams;
 
-		const func = this.page === '/recycle-bin' ? this._handleDeletedVideoSearch : this._handleVideoSearch();
-		func({
+		const searchFunc = this.page === '/recycle-bin' ? this._handleDeletedVideoSearch : this._handleVideoSearch;
+		searchFunc({
 			append,
 			createdAt: dateCreated,
 			query: searchQuery,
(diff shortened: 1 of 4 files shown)


applications/d2l-capture-central/src/components/capture-central-list.js
```diff
@@ -0,0 +1,241 @@
...
+			this.requestUpdate();
+			await this.updateComplete;
+		}
+	}
+
+	async loadNext({ append = true} = {}) {
+		this.loading = true;
+		const { sortQuery, searchQuery, dateModified, dateCreated } = this.queryParams;
+
+		const func = this.page === '/recycle-bin' ? this._handleDeletedVideoSearch : this._handleVideoSearch();
```
AdmiralCoco, line 159: Is `_handleDeletedVideoSearch` to be added? (I didn't find any other references to it from search/this PR)

Also I think you want to set this to `_handleVideoSearch` instead of the result of calling it?


Nit: maybe rename `func` to something like `searchFunc` so it is clear what this is doing?

AdmiralCoco (approved):

applications/d2l-capture-central/src/util/content-service-client.js
```diff
@@ -169,6 +169,36 @@ export default class ContentServiceClient {
...
+	searchDeletedContent({
+		start = 0,
+		size = 15,
+		sort = 'updatedAt:desc',
+		query = '',
+		contentType = '',
+		updatedAt = '',
+		includeThumbnails = false
+	}) {
+		console.log('SEARCHING DELETED CONTENT');
```
MichaelKovalchuk, line 181: do we need to keep this log?

MichaelKovalchuk (approved):

PR #195, opened 2022-01-21
US134918: Rename "My Videos" to "Everyone's Videos" for admin users

MarquisLP:
There is an existing bug in "My Videos": if the user has the "Content Service">"Can Manage All Objects" permission, "My Videos" will display videos from all users.

Instead of creating separate views for "All Videos" and "My Videos", we decided to turn this bug into a feature by renaming the "My Videos" page to "Everyone's Videos" when the "Can Manage All Objects" permission is granted. This means that Admins will no longer have a page showing only their own videos.

So a normal user will see the following in the left nav bar:

- My Videos
- Recycle Bin

My Videos

Whereas a user with "Can Manage All Objects" will see:

- Everyone's Videos
- Recycle Bin

Everyone's Video

(The "Live Events" page will be hidden in a followup PR, as Peter has decided that Live Events won't be on the roadmap yet.)

***

Since the "My Videos" page can now be either "My Videos" or "Everyone's Videos", I've renamed its URL path from `/my-videos` to `/videos` to cover both cases. I've also renamed files and classes to reflect this change.


applications/d2l-capture-central/src/d2l-capture-central-app.js
```diff
@@ -97,15 +105,32 @@ class D2lCaptureCentralApp extends DependencyRequester(NavigationMixin(InternalL
...
 
 	render() {
+		if (this._loading) {
+			return html`
+				<d2l-loading-spinner
+					class="d2l-capture-central-loading-spinner"
+					size="150"
+				></d2l-loading-spinner>
+			`;
+		}
```
MarquisLP, line 133: The Capture Central component's initial setup -- namely, loading user permissions -- can take a few seconds to complete. This can cause race conditions in the UI, such as "Everyone's Videos" briefly appearing as "My Videos".

Because of this, I've made Capture Central render only a loading spinner until the initial setup is complete.

applications/d2l-capture-central/d2l-capture-central.js
```diff
@@ -11,6 +11,7 @@ class D2lCaptureCentral extends DependencyProvider(LitElement) {
 	static get properties() {
 		return {
 			apiEndpoint: { type: String, attribute: 'content-service-endpoint' },
+			canManageAllVideos: { type: Boolean, attribute: 'can-manage-all-videos' },
```
MarquisLP, line 14: Corresponding LMS PR: https://github.com/Brightspace/lms/pull/18134

clemencetbk (approved):

PR #203, opened 2022-01-31
Enforce file size limit

lchan131:
Seee https://trello.com/c/pZrN4cl4/341-video-audio-upload-enforce-upload-limit-in-uploader.


qwangd2l:
I looked at the trello card and is `maxFileSizeInBytes` supposed to be passed in as a property? Right now it's ```this.maxFileSizeInBytes = 1024 * 1024 * 1024 * 5; // 5GB```

lchan131:
> I looked at the trello card and is `maxFileSizeInBytes` supposed to be passed in as a property? Right now it's `this.maxFileSizeInBytes = 1024 * 1024 * 1024 * 5; // 5GB`

Where should we be pulling this value from?

qwangd2l:
> > I looked at the trello card and is `maxFileSizeInBytes` supposed to be passed in as a property? Right now it's `this.maxFileSizeInBytes = 1024 * 1024 * 1024 * 5; // 5GB`
> 
> Where should we be pulling this value from?

I believe Michael wrote some useful information in the trello card regarding this. It'll be passed in from the config variable value from the LE as a property to either this component or the component rendering this one.

core/d2l-content-uploader/src/components/upload.js
```diff
@@ -16,6 +16,7 @@ export class Upload extends RtlMixin(RequesterMixin(InternalLocalizeMixin(LitEle
 	static get properties() {
 		return {
 			errorMessage: { type: String, attribute: 'error-message', reflect: true },
+			maxFileSizeInBytesString: {type: String, attribute: 'max-file-size'},
```
AdmiralCoco, line 19: 1. I think you want to keep this named `maxFileSizeInBytes` (there are other usages of it in the file)
2. You can set the type to `Number`, which should cause lit to parse the property itself, removing the need to `parseInt` on line 117 ([link](https://lit.dev/docs/v1/components/properties/#attributes)) - (although is the `parseInt` required? I would expect it to just pass the value down to be substituted in the string)

AdmiralCoco: Nit: spacing between `{ and }`

core/d2l-content-uploader/src/components/main.js
```diff
@@ -28,6 +28,7 @@ export class Main extends InternalLocalizeMixin(MobxReactionUpdate(ProviderMixin
 			orgUnitId: { type: String, attribute: 'org-unit-id' },
 			tenantId: { type: String, attribute: 'tenant-id' },
 			topicId: { type: String, attribute: 'topic-id' },
+			maxFileUploadSize: { type: Number, attribute: 'max-file-upload-size'},
```
AdmiralCoco, line 31: Nit: spacing before `}`

core/d2l-content-uploader/src/components/upload.js
lchan131: The reason for the name is because it is a string. In the html, it is being passed down as a string, so I can't avoid using parseInt. This is also why maxFileSizeInBytesString exists: To be converted into a number

core/d2l-content-uploader/src/components/main.js
```diff
@@ -28,6 +28,7 @@ export class Main extends InternalLocalizeMixin(MobxReactionUpdate(ProviderMixin
 			orgUnitId: { type: String, attribute: 'org-unit-id' },
 			tenantId: { type: String, attribute: 'tenant-id' },
 			topicId: { type: String, attribute: 'topic-id' },
+			maxFileUploadSize: { type: String, attribute: 'max-file-upload-size'},
```
AdmiralCoco, line 31: Nit: spacing before `}`

core/d2l-content-uploader/src/components/upload.js
```diff
@@ -16,6 +16,7 @@ export class Upload extends RtlMixin(RequesterMixin(InternalLocalizeMixin(LitEle
 	static get properties() {
 		return {
 			errorMessage: { type: String, attribute: 'error-message', reflect: true },
+			maxFileSizeInBytes: {type: Number, attribute: 'max-file-size'},
```
AdmiralCoco, line 19: Nit: spacing between `{ }`

AdmiralCoco (approved):

PR #206, opened 2022-02-09
add disable set preference attribute to Advanced Editing

lchan131:
Add `disable-set-preferences` attribute to `d2l-labs-media-player`


rcausey (approved):
Looks good. Wait until the media-player build finishes and a new version is released. This repo also uses the auto-versioning I discuss there so you will want to do the same thing here, but in this case it's a bug fix (so prefix with `fix` instead of `feat`).

PR #212, opened 2022-02-16
fix: stop page from navigating to videos instead of producer

clemencetbk:
The landing page's `connectedCallBack` is called when the `d2l-capture-central-producer` component is updated. This caused the landing page to appear instead of producer whenever the user clicked 'Edit' after navigating to Capture Central using the nav bar (I think the reason it seemed to work locally was because it didn't cause issue if the nav bar was not clicked, because the landing page component never got created).


applications/d2l-capture-central/src/pages/d2l-capture-central-landing.js
```diff
@@ -16,6 +17,16 @@ class D2lCaptureCentralLanding extends DependencyRequester(navigationMixin(PageV
 
 	connectedCallback() {
 		super.connectedCallback();
+	}
+
+	update() {
```
clemencetbk, line 22: Putting the navigation inside `update` instead of `connectedCallback` so the landing page continues to go to videos after the first update

MichaelKovalchuk (approved):

MarquisLP (approved):

PR #222, opened 2022-02-24
Add 'Please come back later' when latest revision is not ready

clemencetbk:


applications/d2l-capture-central/src/util/content-service-client.js
```diff
@@ -96,6 +96,12 @@ export default class ContentServiceClient {
 		});
 	}
 
+	async getLatestRevision(id) {
+		const content = await this.getContent(id);
+		const reversedCopyOfRevisions = content.revisions.slice().reverse();
+		return reversedCopyOfRevisions.find(revision => !revision.draft);
+	}
+
```
clemencetbk, line 104: The content service  `search ` call only returns ready revisions, so a revision that's still processing won't show up

applications/d2l-capture-central/src/components/videos/content-list-item.js
```diff
@@ -67,8 +67,9 @@ class ContentListItem extends DependencyRequester(navigationMixin(InternalLocali
 		this.apiClient = this.requestDependency('content-service-client');
 	}
 
-	firstUpdated() {
+	async firstUpdated() {
 		super.firstUpdated();
+		if (this.id) this.revisionId = (await this.apiClient.getLatestRevision(this.id)).id;
```
MarquisLP, line 72: Hm, I'm a bit concerned about making the `getLatestRevsion()` request here as part of the list item. This would mean a lot of API requests happening at once on the `/videos` page.

It might be better to make the `/preview` page call `getLatestRevision()` using the given id (and change the route from `/preview/:id/:revisionId` to `/preview/:id` -- I think the preview page would always show the latest revision anyway).

MarquisLP (approved):

PR #232, opened 2022-02-28
feat: Add Launch button and enhance UI in "Capture Encoder" page

MarquisLP:
The "Capture Encoder" page has been redesigned according to designer feedback:

Capture Central -  Capture Encoder  page screenshot

* The "Launch" button opens the custom protocol link `d2lce://launch`, which will open the installed Encoder X app on both Mac and Windows. (The exact behaviour varies depending on browser.)
    * If Encoder X is not installed and the user clicks "Launch", nothing happens. Ideally, we would show a warning message in an alert toast, but [catching custom protocol errors is not straightforward](https://stackoverflow.com/questions/24779312/simplest-cross-browser-check-if-protocol-handler-is-registered). For now, the text instructions pictured above should discourage users from trying to launch before installing.
* The "Download" buttons are automatically reordered depending on the user's OS. (e.g. Windows users would see "Download for Windows" listed first, unlike in the screenshot above.) This is based on the user agent, which isn't 100% reliable, so that's why both download buttons are always available.


applications/d2l-capture-central/locales/en.js
```diff
@@ -12,7 +12,10 @@ export const val = {
 	cancel: 'Cancel',
 	captureCentral: 'Capture Central',
 	captureEncoder: 'Capture Encoder',
+	captureEncoderDescription: 'Capture Encoder is a tool used for recording camera and screen presentations, and publishing them to Brightspace.',
+	captureEncoderLaunchInstructions: 'If this is your first time, please make sure you download and install Capture Encoder to your computer before attempting to launch it from this screen.',
 	captureVideos: 'Capture Videos',
+	downloadForPlatform: 'Download for {platform}',
```
MichaelKovalchuk, line 18: this may not work for Chinese and, maybe, some other languages. 
I'd suggest to use two separate langterms: downloadForWindows, downloadForMac

MichaelKovalchuk: image

MichaelKovalchuk: image

applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js
```diff
@@ -70,8 +71,25 @@ class D2LCaptureCentralEncoder extends InternalLocalizeMixin(DependencyRequester
...
+							class="d2l-capture-central-encoder-button"
+							@click="${this._launchEncoder}"
+							primary
+						>
+							${this.localize('launch')}
+						</d2l-button>
+						${this._encoderDownloads.map(encoderDownload => this._renderEncoderDownloadButton(encoderDownload))}
+					</div>
+					<d2l-alert-toast
+						id="launch-error-alert"
```
MichaelKovalchuk, line 88: if we can not detect that encoder is not installed - we should probably remove this toast for now?

applications/d2l-capture-central/locales/en.js
MarquisLP: Oh, the format of this langterm may be misleading. The raw platform name from the `GET /encoder/update` response isn't what gets passed in -- [it gets converted into another langterm for either Windows or Mac](https://github.com/Brightspace/content-components/blob/855101143bce5164583e48f89f75712981a17076/applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js#L171). So the usage of this langterm would look like:

`${this.localize('downloadForPlatform', { platform: this.localize('windows') })}`

But since Encoder only supports 2 platforms, I can change this to `downloadForWindows` and `downloadForMac` if that would make things simpler.

applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js
MarquisLP: Thanks for catching this -- forgot to remove it after attempting to detect Encoder installation 😅

applications/d2l-capture-central/locales/en.js
MarquisLP: I've gone ahead and condensed the langterms into `downloadForWindows` and `downloadForMac` to make the code clearer.

applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js
```diff
@@ -26,31 +28,30 @@ class D2LCaptureCentralEncoder extends InternalLocalizeMixin(DependencyRequester
...
 	}
 
 	async connectedCallback() {
 		super.connectedCallback();
+
+		if (navigator.userAgent.includes('Macintosh')) {
+			this._userPlatform = 'mac';
+		} else {
+			this._userPlatform = 'win';
+		}
```
ktonon, line 50: What about other platforms? For now, it looks like this is only being used to reorder the download buttons, so I guess it's ok.

applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js
```diff
@@ -83,35 +95,32 @@ class D2LCaptureCentralEncoder extends InternalLocalizeMixin(DependencyRequester
...
-				};
+				return this.localize('downloadForWindows');
 			default:
-				throw new Error(`Invalid platform name: ${shortPlatformName}`);
+				throw new Error(`Invalid platform name: ${platform}`);
+		}
+	}
+
+	_launchEncoder() {
+		window.location.href = 'd2lce://launch';
```
ktonon, line 110: This will change once encoder supports passing in the endpoint and auth token on launch

ktonon (approved):

applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js
MarquisLP: Yeah, this is written under the assumption that Encoder will only support Mac and Windows for the foreseeable future. `this._userPlatform` is indeed only used for ordering the buttons, but defining it here could be misleading as it implies other usages. For clarity, I'll move this block closer to the code that actually uses it.

ktonon: Sounds good. My main concern was that it would be used to say something like "Looks like you're using Windows...", when the user is actually using Linux

ktonon: Would it be worth saying "You have an unsupported platform" in the case that we detect something we don't have a build for?

ktonon: Also, what happens when a mobile browser visits this page?

MichaelKovalchuk: some of our clients do not include user agent in their requests. not sure if it is a big issue or not, but we could detect if it is Mac or Windows and show this, detect if not supported platform - disable launch button and display the question mark with d2l-tooltip message, if it is mobile - do not allow download or launch. Just show the message

MichaelKovalchuk (approved):

applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js
MarquisLP: Hm, what if we just show the message (bolded or with some other emphasis), but still show the Download buttons? That way, just in case the user agent detection gets it wrong, the user can still download. The message could be something like:

> You seem to be using an unsupported platform. Capture Encoder is currently only available for Mac and Windows. If you're on one those platforms, you may still try downloading below.

ktonon: Should we discuss this with Mariangelica?

MarquisLP: As discussed with Michael, mobile platforms will have the launch and download buttons completely hidden and replaced with a message:

Capture Central - Capture Encoder page - Mobile

For unsupported desktop platforms, the launch button will be disabled alongside a message that explains why:

Capture Central - Capture Encoder page - Unsupported Desktop

applications/d2l-capture-central/src/pages/encoder/d2l-capture-central-encoder.js
```diff
@@ -83,35 +113,48 @@ class D2LCaptureCentralEncoder extends InternalLocalizeMixin(DependencyRequester
...
+			// As of iOS 13, the iPad user agent string now contains "Macintosh" instead of "iPad".
+			// To distinguish between Mac and iPad, we need to use the "navigator.standalone" property,
+			// which is only defined in iOS Safari instances.
+			const isMac = RegExp(/Macintosh/i).test(navigator.userAgent);
+			if (isMac && navigator.standalone !== undefined) {
+				isMobile = true;
+			}
+		}
+		return isMobile;
+	}
```
MarquisLP, line 141: This was the best check for mobile platforms that I could come up with, combining several solutions to cover all bases. I decided against screen size-based checks because we aren't concerned about layout sizing/positioning.

MichaelKovalchuk (approved):

package.json
```diff
@@ -53,6 +53,7 @@
     "d2l-resize-aware": "github:BrightspaceUI/resize-aware#semver:^1",
     "enum": "^3.0.4",
     "file-drop-element": "^1.0.1",
+    "ismobilejs": "^1.1.1",
```
MarquisLP, line 56: @MichaelKovalchuk I went with this package over `is-mobile` because the latter doesn't have built-in ES6 support.

PR #233, opened 2022-02-28
Use previewUrl to go to preview from row

clemencetbk:
Setting href makes the page go to `<instance url>/d2l/preview/<content id>` instead of `<instance url>/d2l/wcs/capture-central/<org unit id>/preview/<content id>`

### DIFF
--- a/applications/d2l-capture-central/src/components/videos/content-list-item.js
+++ b/applications/d2l-capture-central/src/components/videos/content-list-item.js
@@ -79,7 +79,8 @@ class ContentListItem extends DependencyRequester(navigationMixin(InternalLocali
 		return html`
 			<d2l-list-item class="d2l-body-compact"
 				?disabled=${this.disabled}
-				href="${previewUrl}"
+				@click=${this._goTo(previewUrl)}
+				href="#"
 				label="${this.title}"
 				?selectable=${this.selectable}
 				key=${this.id}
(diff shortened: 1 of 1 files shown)


applications/d2l-capture-central/src/components/videos/content-list-item.js
```diff
@@ -79,7 +79,7 @@ class ContentListItem extends DependencyRequester(navigationMixin(InternalLocali
 		return html`
 			<d2l-list-item class="d2l-body-compact"
 				?disabled=${this.disabled}
-				href="${previewUrl}"
+				@click=${this._goTo(previewUrl)}
```
MarquisLP, line 82: Is the "..." menu button still clickable? I think I tried this approach previously, but the `<d2l-list-item>`'s clickable area ended up overshadowing everything, so clicking "..." would redirect to preview instead of opening the menu.

clemencetbk: Ah, you're right. I'll try the href approach then

MarquisLP: One thing I didn't consider when using `href` is that it bypasses the `RoutingStore` that manages all of the URL routing in Capture Central. So using it here might not be ideal.

Another approach to try is [using event.stopPropagation()](https://stackoverflow.com/a/1369066) in the `@click` for the "..." menu button. It should stop the `<d2l-list-item>`'s click event from triggering unintentionally.

clemencetbk: That approach works! The only drawback is that when you hover over the row, it doesn't become shaded like it does when href is set... So it doesn't necessarily look clickable

MichaelKovalchuk: can we make it look clickable by adding `href="#"` or it would not work either?

clemencetbk: That works!

MarquisLP (approved):

PR #234, opened 2022-03-01
Allow user to manually edit captions cues timestamps via typing

will-malisch:


capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -193,12 +193,14 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
...
 			this.dispatchEvent(new CustomEvent('media-loaded', { composed: false }));
 		});
+
 	}
 
 	render() {
 		const zoomMultiplierStyleMap = {
 			opacity: this._zoomMultiplierDisplayOpacity
 		};
+		const mediaPlayerDuration = this._mediaPlayer?.duration;
```
MichaelKovalchuk, line 203: i do not think we use this variable

capture/d2l-capture-producer/locales/en.js
```diff
@@ -36,6 +36,8 @@ export const val = {
 	getProcessingProgressError: 'Failed to load processing status for this media file.',
 	hideExpandedCaptionsCueControls: 'Hide expanded controls for this captions cue',
 	insertNewCaptionsAtCurrentTime: 'Insert new captions at current time',
+	invalidFormat: 'HH:MM:SS.sss',
```
MichaelKovalchuk, line 39: How about we us `Time must be in the following format: HH:MM:SS.sss`

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -182,7 +182,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 	firstUpdated() {
 		super.firstUpdated();
 
-		this._mediaPlayer = this.shadowRoot.querySelector('d2l-labs-media-player');
+		this._mediaPlayer = this.shadowRoot.querySelector('d2l-labs-media-player'); // why is this variable getting a this. call, when it is not a property of the class
```
MarquisLP, line 185: > why is this variable getting a this. call, when it is not a property of the class

So in LitElement, a component is allowed to have variables assigned to `this` even if they're not declared as a property in `static get properties()`. In this case, `this._mediaPlayer` is just a normal private variable that can be accessed from within the component. Meanwhile, the properties in `static get properties()` are used by LitElement to determine when to re-render.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -330,6 +333,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 		if (this.enableCutsAndChapters && changedProperties.has('metadata') && this.metadata && this._videoLoaded && this._cutsDifferInMetadataAndTimeline()) {
 			this._resetTimelineWithNewCuts(this.metadata.cuts);
 		}
+		// this.shadowRoot.querySelector('d2l-video-producer-captions').mediaPlayerDuration = this._mediaPlayer.duration;
```
MarquisLP, line 336: Minor: looks like a temporary line for testing that was accidentally kept in.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
MarquisLP: Yeah, looks like it was meant to be used on line 257.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -254,9 +404,17 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
 		`;
 	}
 
+	_setNewTimeVariables(){
+		this.newStartTime = formatTimestampText(this.cue.startTime);
+		this.newEndTime = formatTimestampText(this.cue.endTime);
+	}
+
```
MarquisLP, line 411: Doesn't seem like this function is used anywhere.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
will-malisch: Oh nevermind, I see whats going on here, I didn't realize that I left "why is this variable getting a this. call, when it is not a property of the class" as a comment, my bad.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -153,22 +187,77 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
...
+				detail: {
+					cue: this.cue,
+					endTimeEditedByTextInput: true,
+					newEndTime: unformatTimestampText(this.newEndTime),
+				},
+				bubbles: true,
+				composed: true,
+			}));
+		}
+	}
```
MarquisLP, line 231: It's good that you're using the same event-driven approach for updating the timestamp to keep things consistent! However, using `captions-cue-start/end-timestamp-synced` for both the "Sync to timeline" button and manual timestamp text edits results in a few issues:

- The "synced" in the event name means "synced to timeline position", so the name is misleading when it's used to describe manual timestamp edits via text.
- It adds complexity, as we now need to distinguish between instances of this event that were fired from the "Sync" buttons, and instances that were fired from the timestamp text input. This makes [the event handlers](https://github.com/Brightspace/content-components/blob/d36ef5a3bb2c36f1c6c9816926b8788d00d5f794/capture/d2l-capture-producer/d2l-capture-producer-editor.js#L925) in `d2l-capture-producer-editor.js` more complex as well.

Because of that, I think it would be preferable to **make a new event type** (named something like `captions-cue-start/end-timestamp-edited`) and add new event handlers for them in `d2l-capture-producer-editor.js`.  
`_handleSyncStart/EndTimestampFocusout()` would fire this new event type, while `_handleSyncStart/EndTimestampClicked()` would remain as-is and fire `captions-cue-start/end-timestamp-synced`.

capture/d2l-capture-producer/src/captions-utils.js
```diff
@@ -80,6 +80,30 @@ function formatTimestampText(timestampInSeconds) {
 	return `${hoursMinutesSecondsString}.${millisecondsString}`;
 }
 
+/**
+ * Takes a string of and validates that is is of the format hh:mm:ss.sss, hh:mm:ss.ss, hh:mm:ss.s, hh:mm:ss., or hh:mm:ss
+ * @param {string} timestampInText A timestamp string in format described above, storing hours, minutes, and seconds to three or less decimal places
+ * @returns true if the string is correctly formated, and false if it is not
+ */
+function validTimestampFormat(timestampInText) { // Timestamp format from function formatTimestampText(timestampInSeconds): 00:00:00.000
+	const format = /[0-9][0-9]:[0-9][0-9]:[0-9][0-9](\.[0-9][0-9][0-9]|\.[0-9][0-9]|\.[0-9]|\.|)$/;
```
MarquisLP, line 89: It may be possible to shorten this regex and reduce repetition. Some resources:

- https://www.regextutorial.org/regex-for-numbers-and-ranges.php
- https://stackoverflow.com/questions/37194527/how-do-i-repeat-regex

capture/d2l-capture-producer/test/captions-utils.test.js
```diff
@@ -28,6 +28,80 @@ describe('captions-utils.js', () => {
...
+		});
+
+		it('works for timestamps that contain milliseconds', () => {
+			const expected = 6065.456;
+			const actual = unformatTimestampText('01:41:05.456');
+			assert.equal(actual, expected);
+		});
+	});
+
+	describe('validTimestampFormat', () => {
```
MarquisLP, line 57: Nice job covering all these cases! Can we add a few more for these edge cases:

- Empty string
- Garbage text that doesn't resemble the format at all (e.g. "asdfghjkl123")

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -195,21 +286,79 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
...
+					id="d2l-advanced-editing-disabled-icon-start"
+					icon="tier1:help"
+				></d2l-icon>
+				<d2l-tooltip
+					id="start-time-tooltip"
+					for="d2l-advanced-editing-disabled-icon-start"
+					state="error"
+					showing
+				>${this.localize(this.errors[this.startValidationError], { formattedTime: formatTimestampText(this.mediaPlayerDuration)})}</d2l-tooltip>`
+			: null;
```
MarquisLP, line 302: Would it be possible to put the tooltip around the timestamp text box itself, instead of this new icon? From what I've seen on various sites, having the error tooltip attached to the textbox is the usual behaviour. The `<d2l-tooltip>` docs show this as an example:

</img>

MarquisLP (changes requested):

MarquisLP:
Some more things that Will and I noticed that should be fixed/addressed:

- In certain cases, the Start/End timestamps will revert back to their original value when losing focus. Since we're using the "?" icon with tooltip approach, we shouldn't be reverting the timestamp values (except possibly when saving -- see the next point).
- If the user Saves or Finishes when there are errors in 1 or more timestamps, those timestamp values don't get reverted back to their original value in the UI. This technically isn't a critical issue, as the original values are what actually get saved, but it might mislead users into thinking that the errors got saved.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
MarquisLP: Discussed this with Will and we decided to stay with the "?" icon + tooltip approach. If we were to put the tooltip on the textbox and have it automatically disappear, then the user might never get to see the tooltip (e.g. if they scroll down to another cue off-screen and change focus to that cue).

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -101,34 +110,59 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
...
-			>
-				${this._renderMainControls()}
-				${this.expanded ? this._renderExpandedControls() : ''}
-			</div>
+		<div
+		class="d2l-video-producer-captions-cues-list-item${this.active ? '-active' : ''}"
+		>
+		${this._renderMainControls()}
+		${this.expanded ? this._renderExpandedControls() : ''}
+		</div>
```
MarquisLP, line 152: Looks like the indentation got mixed up here 😅 Can you add back the indentation?

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -101,34 +110,59 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
 				grid-column: 1;
 			}
 
-			.d2l-video-producer-captions-cue-timestamp-text {
+			.d2l-video-producer-captions-cue-timestamp-input {
 				grid-column: 2;
 			}
```
MarquisLP, line 115: Were you able to shrink the width of the timestamp text input?

MarquisLP (approved):
LGTM! Just had a few minor comments.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -962,6 +967,43 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 		}));
 	}
 
+	_handleChangeCaptionsCueEndTimestampByTextInput(event) {
```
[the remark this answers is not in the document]
will-malisch: Will do.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: Yes, I centered it and reduced this width a touch

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: fixed

capture/d2l-capture-producer/src/captions-utils.js
```diff
@@ -80,6 +80,29 @@ function formatTimestampText(timestampInSeconds) {
 	return `${hoursMinutesSecondsString}.${millisecondsString}`;
 }
 
+/**
+ * Takes a string of and validates that is is of the format hh:mm:ss.sss, hh:mm:ss.ss, hh:mm:ss.s, hh:mm:ss., or hh:mm:ss
+ * @param {string} timestampInText A timestamp string in format described above, storing hours, minutes, and seconds to three or less decimal places
+ * @returns true if the string is correctly formated, and false if it is not
+ */
+function validTimestampFormat(timestampInText) { // Timestamp format from function formatTimestampText(timestampInSeconds): 00:00:00.000
+	const format = /([\d]{2}:){2}[\d]{2}(\.[\d]{3}|\.[\d]{2}|\.[\d]|\.|)$/;
```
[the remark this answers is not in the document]
will-malisch: thanks!

capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -535,6 +535,7 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 				this._metadataChanged = false;
 			}
 			if (this._captionsChanged) {
+				this._captionsLoading = true;
```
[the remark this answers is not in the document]
will-malisch: fixed

capture/d2l-capture-producer/d2l-capture-producer-editor.js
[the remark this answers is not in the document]
will-malisch: Done.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -967,14 +930,22 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
...
+			const newEndTime = event.detail.newEndTime;
+			editedCue.endTime = Math.floor(newEndTime * 1000) / 1000; // WebVTT uses 3-decimal precision.
+			if (editedCue.endTime <= editedCue.startTime) {
+				editedCue.startTime = Math.max(editedCue.endTime - cueDuration, 0);
+			}
+		} else {
+			editedCue.endTime = Math.floor(this._mediaPlayer.currentTime * 1000) / 1000; // WebVTT uses 3-decimal precision.
+			if (editedCue.endTime <= editedCue.startTime) {
+				editedCue.startTime = Math.max(editedCue.endTime - cueDuration, 0);
+			}
```
MarquisLP, line 948: Looks good! I think we can condense this a bit further and remove the if-else blocks, as both blocks are pretty similar.

```
const newEndTime = event.detail.newEndTime ?? this._mediaPlayer.currentTime
editedCue.endTime = Math.floor(newEndTime * 1000) / 1000;
...
```

capture/d2l-capture-producer/d2l-capture-producer.js
```diff
@@ -535,7 +535,7 @@ class CaptureProducer extends RtlMixin(InternalLocalizeMixin(LitElement)) {
 				this._metadataChanged = false;
 			}
 			if (this._captionsChanged) {
-				this._captionsLoading = true;
+				this._captionsLoading = true; // Force the captions tab to reload in order to reset any invalid timestamp text inputs.
```
MarquisLP, line 538: 👍

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -112,6 +113,9 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
 
 			.d2l-video-producer-captions-cue-timestamp-input {
 				grid-column: 2;
+				margin-left: auto;
+				margin-right: auto;
+				width: 90%;
 			}
```
MarquisLP, line 119: Nice, looks like this will work. Just so Michael and I can see what this looks like now, can you post a screenshot?

MarquisLP (approved):
Looks ready to me! I just had one minor comment for refactoring the code a little bit. Once you push a fix for that, you can feel free to go ahead and merge.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: Screen Shot 2022-03-04 at 12 39 20 PM

will-malisch: hows that?

MarquisLP: Looks good! It matches the original positioning quite closely, which is nice.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
will-malisch: Okay

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: Sweet, thanks

MarquisLP (approved):
LGTM, ready to merge 👍

PR #237, opened 2022-03-07
Updating encoder launch endpoint to provide tenant id

ktonon:
launch


ktonon:
I haven't tested this locally. How is that done?

ktonon:
Tried testing locally, but the capture central link was still going to what looks like the old version. Complaining about missing lti secret and portal url. I'm updating my lms and doing a full_all now...

ktonon:
Think I'm missing a feature flag

ktonon:
I was missing `d2l.Tools.Capture.UseContentService` to `on`

ktonon:
Tested locally and the encoder was launched with the correct remote command

gordon-plumb (approved):

PR #238, opened 2022-03-07
feat: Bump version for alerting user on timestamp text inputs in Producer

will-malisch:


capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -289,30 +291,18 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
...
 			html`
-				<d2l-icon
-					class="start-timestamp-row"
-					id="d2l-advanced-editing-disabled-icon-start"
-					icon="tier1:help"
-				></d2l-icon>
 				<d2l-tooltip
 					id="start-time-tooltip"
-					for="d2l-advanced-editing-disabled-icon-start"
+					for="start-${this._startInputId}"
```
MichaelKovalchuk, line 296: we do not need to prefix it with the `start` or `end`
see: https://search.d2l.dev/xref/BrightspaceUI/core/components/count-badge/count-badge.js?r=641777a1#43

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -321,20 +311,21 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
 			<div class="d2l-video-producer-captions-cue-expanded-controls">
 				<div class="d2l-video-producer-captions-cue-start-end-grid">
 					<p class="start-timestamp-row d2l-label-text d2l-video-producer-captions-cue-timestamp-label">${this.localize('captionsCueStartTimestamp')}</p>
-					<p class="start-timestamp-row d2l-body-standard d2l-video-producer-captions-cue-timestamp-input">
+					<p class="start-timestamp-row d2l-body-standard d2l-video-producer-captions-cue-timestamp-input" id="d2l-video-producer-captions-cue-start-timestamp-input">
```
MichaelKovalchuk, line 314: do we need this id? d2l-video-producer-captions-cue-start-timestamp-input

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: My bad, I meant to delete that.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: also meant to delete

MarquisLP:
Since this PR will have multiple commits, here's a tip to make sure a new release gets created on merge:

On the button that allows you to merge, there's a dropdown with some merge options. Pick "Squash and merge" to combine all your changes into a single commit. Then, it will ask you for the title and description of this single commit. As long as you set title to `feat: <your title here>`, the release should be created as soon as you merge.

Merge Options

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -321,20 +311,21 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
...
 						<d2l-input-text
-							id="start-time"
+							id="start-${this._startInputId}"
 							class="d2l-input-text start-time-input-text"
 							label="Cue Start Time"
 							label-hidden
+							aria-invalid="${this.startValidationError !== ''}"
```
MarquisLP, line 320: Nice catch with this accessibility property 👍 

Nitpick: for HTML attributes, we prefer to keep them in alphabetical order. So `aria-invalid` should be above `class`, `id` should be above `@input`, and so on.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -474,7 +466,6 @@ class VideoProducerCaptions extends InternalLocalizeMixin(LitElement) {
 				overflow-x: hidden;
 				overflow-y: scroll;
 				padding: 0px 10px 10px 0px;
-				position: relative;
```
MarquisLP, line 477: For future reference: removing `position: relative` here fixes a bug that caused timestamp error tooltips to appear in the wrong location.

@will-malisch: if you know more about why `position: relative` caused the bug, can you share it here? This is just so we have an explanation here for anyone looking at this PR in the future.

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
```diff
@@ -289,30 +291,18 @@ class CaptionsCueListItem extends InternalLocalizeMixin(LitElement) {
...
 				>${this.localize(this.errors[this.startValidationError], { formattedTime: formatTimestampText(this.mediaPlayerDuration)})}</d2l-tooltip>`
 			: null;
 
 		const endTimeTooltip = this.endValidationError ?
 			html`
-				<d2l-icon
-					class="end-timestamp-row"
-					id="d2l-advanced-editing-disabled-icon-end"
-					icon="tier1:help"
-				></d2l-icon>
```
MarquisLP, line 311: So it looks like we're removing the ? icon for errors? Can you share a screenshot of what the list items look like now?

(Tip: for any PR where you make UI changes, it's helpful to the reviewers if you provide screenshots and/or a screen recording.)

MarquisLP:
Can we change the PR title to something like:

"fix: Fix timestamp error tooltips appearing the wrong location"

This is just to make it clear what the purpose of this PR was, for anyone browsing through the PR list or commit history.

Also, `fix:` should be used instead of `feat:` in this case because it's a bug fix. (`fix:`, like `feat:`, will also trigger a new release.)

You can use this same title for the merge commit.

MarquisLP:
My bad, I didn't realize that you were changing this to use the error alert built into `d2l-input-text` (which is really neat!). A better title would be:

"feat: Display timestamp error tooltips inside text input"

Can you edit the first comment of this PR with a short description of this change, as well as a screenshot?

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: OKay, next time I will leave pictures.


Screen Shot 2022-03-08 at 9 27 07 AM
Screen Shot 2022-03-08 at 9 27 13 AM

MarquisLP (approved):

capture/d2l-capture-producer/src/d2l-video-producer-captions.js
will-malisch: Link to thread where Margaree Peacocke helped me: https://d2l.slack.com/archives/C0PHG3QB0/p1646678116312399

Her last response was:
I don't totally know what's going on but I do think that it's related to the position: relative and overflow scrolling. In tooltip when it's positioning itself, it looks for this offset parent (https://github.com/BrightspaceUI/core/blob/main/components/tooltip/tooltip.js#L498). When the .d2l-video-producer-captions-cues-list has position: relative then it positions itself relative to that element. But when that position: relative is removed then the offset parent that it positions itself related to is .d2l-video-producer-captions (2 divs above) which does not have the scrolling affecting it.
[tooltip.js](https://github.com/BrightspaceUI/core/blob/main/components/tooltip/tooltip.js)
        const offsetParent = getOffsetParent(this);

MarquisLP: Thanks! Makes sense that combining relative positioning and scrolling would cause some funky behaviour with the tooltip's offset.

PR #241, opened 2022-03-09
feat: Bump version for editing timestamp text inputs in Producer

will-malisch:
I was reviewing the last update in staging and noticed there was one error that somehow got missed. I honestly think I might have clicked control Z or something because I remember going over this on call with Mark. 

It looks like I copy and pasted a line of code in start time focus out to end time focus out and forgot to change a variable name.


MarquisLP (approved):
LGTM

> I honestly think I might have clicked control Z or something because I remember going over this on call with Mark.

No worries! Easy to miss stuff like this when copying and pasting.

PR #242, opened 2022-03-09
add check for supported file extension

gordon-plumb:


applications/d2l-capture-central/src/pages/videos/d2l-capture-central-videos.js
```diff
@@ -152,15 +152,15 @@ class D2LCaptureCentralVideos extends contentSearchMixin(DependencyRequester(Pag
...
-					errorToastElement.setAttribute('open', true);
-				}
+				this._showUploadErrorToast(this.localize(
+					'fileTooLarge',
+					{ localizedMaxFileSize: formatFileSize(maxFileSizeInBytes) }
+				));
+				event.target.value = '';
+				return;
+			}
+			if (!isSupported(file.name)) {
```
MichaelKovalchuk, line 162: would it also allow you to upload Audio? we are planning to include audio support at some point, but do not want to allow it now

MichaelKovalchuk (approved):

applications/d2l-capture-central/src/pages/videos/d2l-capture-central-videos.js
gordon-plumb: This won't allow audio. We would have to add the supported audio extensions similar to how it is done in d2l-content-uploader: https://github.com/Brightspace/content-components/blob/master/core/d2l-content-uploader/src/util/media-type-util.js#L9

PR #243, opened 2022-03-09
fix: Move language selector button/drop menu to tabs

will-malisch:
Moved the language selector to the tabs section (currently only tabs are chapters and captions), because these were the main components that were affected by the language change.


capture/d2l-capture-producer/src/constants.js
```diff
@@ -105,5 +105,9 @@ const constants = {
 		ZOOM_HANDLE_SET: '#006FBF',
 		ZOOM_HANDLE_UNSET: '#CDD5DC',
 	},
+
+	// Tabs
+	TABS_BODY_HEIGHT: 484,
+	TABS_CONTAINER_HEIGHT: 375,
```
MarquisLP, line 111: Nice, this a great code improvement!

MarquisLP (commented):
LGTM -- don't see any changes that need to be made 🙂 But in keeping with Michael's comment, I won't approve yet.

will-malisch:
Both these were discussed, decided to go with the language selection drop-down button on the right.
Screen Shot 2022-03-09 at 3 20 46 PM
Screen Shot 2022-03-09 at 3 21 43 PM

MarquisLP:
> Both these were discussed, decided to go with the language selection drop-down button on the right.

Sounds good. I prefer having it on the right as well, as having it on the left makes the text look misaligned with "Table of Contents" below.

MichaelKovalchuk:
April dev done is not over yet, we need to wait for one more week before you merge this PR

MichaelKovalchuk (approved):

PR #248, opened 2022-03-10
Add d2l-content-media-player component

AdmiralCoco:


core/d2l-content-media-player/src/clients/rest-client.js
```diff
@@ -0,0 +1,136 @@
...
+			path: `/api/${this.tenantId}/content/${this.contentId}/revisions/${this.revisionId}`
+		});
+		return revision ? this._formatRevision(revision) : null;
+	}
+
+	getThumbnails() {
+		return this._fetch({
+			path: `/api/${this.tenantId}/content/${this.contentId}/revisions/${this.revisionId}/resources/thumbnails/signed-url`,
+			query: {
+				contextId: this.contextId,
```
qwangd2l, line 76: I'm wondering if this is needed since they get added by default in the query if it's a GET request.

qwangd2l, line 76: I'm wondering if this is needed since they get added by default in the query if it's a GET request.

AdmiralCoco: Whoops, you're correct - I left this in when testing

qwangd2l (approved):

PR #250, opened 2022-03-10
Add d2l-content-topic-renderer component

AdmiralCoco:


core/d2l-content-topic-renderer/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+```html
+<script type="module">
+    import '@brightspace/content-components/d2l-content-topic-renderer.js';
+</script>
+<d2l-content-topic-renderer
+	?allow-download=false
+	?allow-download-on-error=false
+	content-service-endpoint="http://example.org"
+	topic-id="12345">
```
qwangd2l, line 23: Should d2lrn also be here in the README?
Or is it not required for this component?

AdmiralCoco: Good question - the D2LRN isn't needed in this component as it [uses the topicId to figure it out](https://github.com/Brightspace/content-components/pull/250/files#diff-aba0f29b517e33fa560392258765ade9b3760b1ad482fa4a2bfb0f2066d4fcaeR18) - I will make the `d2lrn` property private and prefix it with and '_'

qwangd2l (approved):

PR #259, opened 2022-03-17
Timeline components

lchan131:
Add attribute to adjust the size of the width of the timeline.

### DIFF
--- a/capture/d2l-capture-producer/d2l-capture-producer-editor.js
+++ b/capture/d2l-capture-producer/d2l-capture-producer-editor.js
@@ -31,6 +31,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement)) 
 			captions: { type: Array },
 			captionsLoading: { type: Boolean, attribute: 'captions-loading' },
 			captionsUrl: { type: String },
+			canvasWidth: { type: Number, attribute: 'canvas-width'},
 			defaultLanguage: { type: Object },
 			enableCutsAndChapters: { type: Boolean },
 			format: { type: String },
@@ -180,6 +181,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement)) 
 	}
 
 	firstUpdated() {
+		console.log(this.canvasWidth);
 		super.firstUpdated();
 
 		this._mediaPlayer = this.shadowRoot.querySelector('d2l-labs-media-player');
@@ -194,6 +196,10 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement)) 
 			this.dispatchEvent(new CustomEvent('media-loaded', { composed: false }));
 		});
 
+	}
+
+	get _timelineWidth() {
+		return this.canvasWidth - constants.TIMELINE_OFFSET_X * 2;
 	}
 
 	render() {
@@ -272,7 +278,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement)) 
 				${(this.enableCutsAndChapters ? html`
 					<div class="d2l-video-producer-timeline" style="visibility: ${this.timelineVisible ? 'visible' : 'hidden'};">
 						<div id="canvas-container">
-							<canvas height="${constants.CANVAS_HEIGHT}px" width="${constants.CANVAS_WIDTH}px" id="timeline-canvas"></canvas>
+							<canvas height="${constants.CANVAS_HEIGHT}px" width="${this.canvasWidth}px" id="timeline-canvas"></canvas>
 							<div id="zoom-multiplier" style=${styleMap(zoomMultiplierStyleMap)}>
 								${this._getZoomMultiplierDisplay()}
 							</div>
@@ -397,7 +403,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement)) 
 		});
 
 		displayObject.on('pressmove', event => {
-			const pixelsAlongTimelineToMoveTo = CaptureProducerEditor._getPixelsAlongTimelineFromStageX(event.stageX);
+			const pixelsAlongTimelineToMoveTo = this._getPixelsAlongTimelineFromStageX(event.stageX);
 
 			const returnValue = mark.move(pixelsAlongTimelineToMoveTo);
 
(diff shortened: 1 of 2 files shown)


capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -31,6 +31,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 			captions: { type: Array },
 			captionsLoading: { type: Boolean, attribute: 'captions-loading' },
 			captionsUrl: { type: String },
+			canvasWidth: { type: Number, attribute: 'canvas-width'},
```
rcausey: I don't think this makes sense as a property on the editor. How about we just remove it and use the constant for now?

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -518,13 +440,14 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 
 	//#region Timeline management
 	_configureStage() {
-		this._timelineCanvas = this.shadowRoot.querySelector('#timeline-canvas');
```
rcausey, line 521: I think we can move a lot more of this functionality to the timeline component (including the buttons to seek/mark/cut). Ideally, the timeline component UI would work 100% on its own - i.e. you could create a demo page with just this component and use it to create/move/delete marks, create/delete cuts, zoom in/out, seek, and the state of the UI would be handled completely within the component. You could then integrate it into the editor or even some other tool. The editor itself would know nothing about the inner workings of the timeline component and just interact with it via its properties and the events it emits.

rcausey (dismissed):

capture/d2l-capture-producer/d2l-capture-producer-editor.js
rcausey: I would actually suggest creating a demo page for just the timeline or modifying the current demo page to include a section for the timeline component without the surrounding editor).

lchan131: I have separated much of the functionality to the timeline component.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -46,8 +44,8 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 			timelineVisible: { type: Boolean, attribute: 'timeline-visible' },
 
 			_activeCue: { type: Object, attribute: false },
+			_chaptersComponent: { type: Object, attribute: false},
 			_zoomMultiplier: { type: Number, attribute: false },
```
rcausey, line 48: Do we use this for anything other than setting the initial value? We can probably get rid of it and just set the initial value in the component.

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -288,346 +196,58 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
...
-									/>
-								<label for="d2l-video-producer-cut-button" @click="${this._changeToCutMode}" id="d2l-video-producer-cut-button-label">
-									<d2l-icon id="d2l-video-producer-cut-button-icon" icon="html-editor:cut"></d2l-icon>
-									<d2l-tooltip for="d2l-video-producer-cut-button-label" delay="500">${this.localize(constants.CONTROL_MODES.CUT)}</d2l-tooltip>
-								</label>
-							</div>
-						</div>
-					</div>
+					<d2l-capture-producer-timeline
+						.chaptersComponent=${this._chaptersComponent}
```
rcausey, line 200: If we pass in the chapters and media player components as properties, we're making the timeline component very dependent on/tightly coupled to these components. Could we instead handle this via events/properties? 

For instance, the timeline could emit an event when the content marker is moved, and the editor would handle that event and call the setChapterToTime method on the chapters component (instead of the timeline calling that method on the component directly).

Similarly, we could bind media player properties (e.g. duration, currentTime) to properties in the timeline component so that it's aware of updates to the media player properties it cares about without actually directly relying on the media player component.

rcausey (changes requested):
Really nice progress, I think we're getting close

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -78,65 +76,12 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
...
-
-			.d2l-video-producer-timeline-mode-button input[type="radio"]:checked + label {
-				background-color: var(--d2l-color-gypsum);
-			}
-
-			#timeline-canvas {
-				border: ${constants.CANVAS_BORDER_WIDTH}px solid #787878;
-			}
-
 			#zoom-handle {
```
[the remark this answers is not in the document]
lchan131: Weird. This was here originally and it isn't used either. I'll remove it.

MarquisLP (commented):
Looks great for the most part! I just had a few comments about code style, but the core changes are good.

My only question is: do we have a rough idea yet of how we'll handle the next step of making the timeline responsive? I see that the timeline width can now be controlled by passing a `width` property down to the `<d2l-capture-producer-timeline>` component, so I'm curious to know how that property will be used for responsiveness.

EDIT: Just saw your earlier comment about the timeline width, and it's fine if the plan is undecided yet. I was just curious 🙂

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -46,8 +44,7 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
 			timelineVisible: { type: Boolean, attribute: 'timeline-visible' },
 
 			_activeCue: { type: Object, attribute: false },
-			_zoomMultiplier: { type: Number, attribute: false },
-			_zoomMultiplierDisplayOpacity: { type: Number, attribute: false },
+			_chaptersComponent: { type: Object, attribute: false},
```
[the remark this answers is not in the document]
lchan131: You are right. This is not used anymore given the changes I have currently.

rcausey (changes requested):
Just a couple small things

capture/d2l-capture-producer/d2l-capture-producer-editor.js
```diff
@@ -986,24 +364,26 @@ class CaptureProducerEditor extends RtlMixin(InternalLocalizeMixin(LitElement))
...
 	}
-
 	_handleCueChange() {
 		this._activeCue = this._mediaPlayer.activeCue;
 	}
-
 	_handleMediaError() {
 		this.dispatchEvent(new CustomEvent('media-error', { composed: false }));
 	}
-
```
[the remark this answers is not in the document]
lchan131: I tried adding a linting rule for this, but it is changing a lot of other files I never worked on. It might be more appropriate to make a separate PR for this. What do you think?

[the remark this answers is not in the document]
rcausey: Sounds good

rcausey (approved):

MichaelKovalchuk (approved):

PR #260, opened 2022-03-18
Make preview dialog height shorter

clemencetbk:
This is what the preview looks like with shorter height:
image
image


MarquisLP:
Looks much better! Does it look fine on mobile as well (no cutoffs or warped dimensions)?

clemencetbk:
Hmmm, everything works ok on mobile but the preview height is most of the page...
image

MarquisLP:
I see... when these changes are reverted, does the dialog still look like that on mobile?

clemencetbk:
Looks the same without this commit, I guess it would be a separate issue then

PR #262, opened 2022-03-22
add labels for d2l-list-item

gordon-plumb:


gordon-plumb:
Not needed, see: https://github.com/Brightspace/content-components/pull/264

gordon-plumb:
Not needed, see: https://github.com/Brightspace/content-components/pull/264

PR #263, opened 2022-03-22
feat: Display video posters in Capture Central video lists

MarquisLP:
This PR displays video posters in Capture Central's "My/Everyone's Videos" and "Recycle Bin" pages. If a video poster can't be retrieved (e.g. the network request fails), we fall back to showing a video icon like before.

**Everyone's Videos (admin):**

Capture Central - Thumbnails - Admin

I had to limit the poster size to 64px, otherwise the values in the "Owner" and "Modified" columns become too misaligned with the column headers.

Also, I tried to reduce the margin around the thumbnails, but this is controlled by the parent component, `<d2l-list-item>`. The attribute that controls it is `padding-style`, which has 3 options:

- **default**
- **slim (the one used in this PR)** - slightly less margin than default
- **none** - completely removes the horizontal separator line, so I didn't go with this one

**My Videos (instructor/student)**:

Capture Central - Thumbnails - Instructor

**Recycle Bin**:

Capture Central - Thumbnails - Recycle Bin

**Mobile:**

Capture Central - Thumbnails - Mobile


applications/d2l-capture-central/src/components/capture-central-list.js
```diff
@@ -85,10 +85,11 @@ export class CaptureCentralList extends DependencyRequester(InternalLocalizeMixi
 		const { id: revisionId, type } = item.revision;
 		const { id } = item.content;
 		const { ownerDisplayName } = item;
+		const { thumbnail } = item;
```
MichaelKovalchuk, line 88: can we do `const { ownerDisplayName, thumbnail } = item`?

applications/d2l-capture-central/src/components/recycle-bin/recycle-bin-list.js
```diff
@@ -108,12 +108,12 @@ class RecycleBinList extends CaptureCentralList {
 		<recycle-bin-item
 			id=${item.id}
 			revision-id=${item.revisionId}
+			thumbnail=${ifDefined(item.thumbnail)}
```
MichaelKovalchuk, line 111: to not mix with the timeline thumbnail, I think we should rename thumbnail to poster

MichaelKovalchuk (approved):
Looks good

will-malisch (approved):
Looks great Mark

applications/d2l-capture-central/src/components/recycle-bin/recycle-bin-list.js
MarquisLP: Good idea, I'll rename it here and in related files.

clemencetbk (approved):

PR #264, opened 2022-03-23
exclude d2l-content-store from tests

gordon-plumb:
d2l-content-store tests, specifically content-list.test.js and content-list-item.test.js were causing other tests to fail with `Error: Uncaught Error: LabelledMixin: "d2l-list-item" is missing a required "label" attribute (node_modules/@brightspace-ui/core/mixins/labelled-mixin.js:208)` because the error is thrown after a delay of 3 seconds. Since we don't use d2l-content-store, excluding these tests to prevent them from causing failures in other tests.

Other places where we use `d2l-list-item` have the `label` attribute or do not require it.

### DIFF
--- a/web-test-runner.config.js
+++ b/web-test-runner.config.js
@@ -1,7 +1,10 @@
 import { playwrightLauncher } from '@web/test-runner-playwright';
 
 function getPattern(type) {
-	return `+(applications|core|capture)/**/*.${type}.js`;
+	return [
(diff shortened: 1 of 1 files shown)


MarquisLP (approved):
Thanks so much for tracking down the cause of the intermittent CI failures!

I wonder how large of a task it would be to resolve those errors, but that can wait until later since content-store is currently dormant. This should be fine to merge for now.

MichaelKovalchuk (approved):

PR #268, opened 2022-03-25
fix: Stop capture central from rerendering page when opening preview

will-malisch:
In capture central, if you scrolled to the bottom of the My Videos/Everyones Videos List, and clicked on an item to open its preview, the page would re-render, so that the preview would open and behind the preview would be the top of the page.

https://user-images.githubusercontent.com/99415792/160131879-84e42813-968f-4971-b54e-1f869a1f0bf6.mov

 Desired behaviour is for the preview to render at the current location in the list.

https://user-images.githubusercontent.com/99415792/160131958-acd67bf3-6da8-4d7e-bb12-9faf93552c8c.mov


will-malisch:
I am talking to web-platform right now. After merging, we might have to re-visit this later. They are trying to help us find a way to make this a d2l-list-item-button instead of a d2l-list-item. Apparently, there are accessibility issues when using the href property of d2l-list-item for non-sighted users.

clemencetbk (approved):

will-malisch:
Regarding my comment about accessibility, this issue has been resolved. By using a `d2l-list-item-button` instead of `d2l-list-item`, we are able to get the same UI without using the `href` property of `d2l-list-item`, making it more accessible to non-sighted users, because the href would announce to screen reader users as a link, when in fact there is no new link, but rather a dialog.

MichaelKovalchuk (approved):

PR #270, opened 2022-03-28
EsLint spaces between methods

lchan131:
Add a rule for spaces between methods


lchan131:
Oops. I think I branched off of another branch and not the master branch.

PR #272, opened 2022-03-28
Make timeline responsive to screen width change

lchan131:
Demo:

https://user-images.githubusercontent.com/98043376/160444346-2f054211-60aa-45e2-aa1f-9b59a3147823.mov


MichaelKovalchuk:
can we make canvas responsive instead and get its width from the javascript inside the timeline component? Some example: https://stackoverflow.com/questions/34772957/how-to-make-canvas-responsive

rcausey:
> > can we make canvas responsive instead and get its width from the javascript inside the timeline component? Some example: https://stackoverflow.com/questions/34772957/how-to-make-canvas-responsive
> 
> Doing it this way breaks some of the functionality of the timeline. I would just be forced to make a window resize handler anyways, which defeats the whole purpose of making it responsive.
> 
> Do you want the width of the timeline to be independent of the editor?

I think the goal is to make the timeline component behave like other standard tags, like <div> - i.e. you can set a static width on it (e.g. 500px) or you can set width to be some percentage (e.g. 100%), which means it would change based on the size of the parent container. It might be easier to test the timeline component on its own, and testing something like:

```
<d2l-capture-producer-timeline width="100%" ...other params...></d2l-capture-producer-timeline>
```

If you did the above, it should shrink/grow based on the size of the parent container. If the parent container is the body/window, then it would shrink/grow based on the size of the window with no extra code.

lchan131:
Going to do a clean branch

PR #275, opened 2022-03-29
Rename Capture Encoder to Create Presentation

MichaelKovalchuk:


applications/d2l-capture-central/locales/en.js
```diff
@@ -11,7 +11,7 @@ export const val = {
 	cameraAndScreen: 'Camera and screen',
 	cancel: 'Cancel',
 	captureCentral: 'Capture Central',
-	captureEncoder: 'Capture Encoder',
+	captureEncoder: 'Create Presentation',
```
MarquisLP, line 14: Should we rename the variable name as well? `captureEncoder` --> `createPresentation`

MichaelKovalchuk: probably not now, want to keep it consistent with the rest of the captureEncoder variables

MarquisLP (approved):

MarquisLP:
Oh, I think the merge commit was missing a `feat:` prefix, so a new release wasn't created.

PR #286, opened 2022-04-04
Bug fixes and adding additional error messages.

qwangd2l:
Also add offscreen message about title and description of the media resource.

### DIFF
--- a/core/d2l-content-media-player/src/clients/rest-client.js
+++ b/core/d2l-content-media-player/src/clients/rest-client.js
@@ -66,8 +66,7 @@ export default class ContentServiceClient {
 		const revision = await this._fetch({
 			path: `/api/${this.tenantId}/content/${this.contentId}/revisions/${this.revisionId}`,
 			query: {
-				includeTitle: true,
-				includeDescription: true
+				includeTitleAndDescription: true
 			}
 		});
 		return revision ? this._formatRevision(revision) : null;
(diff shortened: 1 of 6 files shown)


core/d2l-content-media-player/src/clients/rest-client.js
```diff
@@ -64,7 +64,11 @@ export default class ContentServiceClient {
 
 	async getRevision() {
 		const revision = await this._fetch({
-			path: `/api/${this.tenantId}/content/${this.contentId}/revisions/${this.revisionId}`
+			path: `/api/${this.tenantId}/content/${this.contentId}/revisions/${this.revisionId}`,
+			query: {
+				includeTitle: true,
```
qwangd2l, line 69: These two query parameters are not expected by content service just yet.
I plan to merge changes in d2l-content-service to return this information when retrieving revisions.

The purpose is to use it for an offscreen message about the media resource like we do for video notes.

rcausey: Do you think we need this much granularity, or could we just have one parameter for `includeTitleAndDescription`? Or even just always return title and description (assuming it can be backwards compatible and doesn't incur a performance hit)?

qwangd2l: I think I would prefer to have one query parameter for both over always returning it, for now.

rcausey (approved):

PR #290, opened 2022-04-07
fix: changes to allow uploading to be compatible with smart-curriculum

willisching:


core/d2l-content-uploader/src/util/content-service-client.js
```diff
@@ -3,15 +3,14 @@ import auth from 'd2l-fetch-auth/src/unframed/index.js';
...
-d2lfetch.use({ name: 'auth', fn: auth });
-
 export default class ContentServiceClient {
 	constructor({
 		endpoint,
 		tenantId
 	}) {
 		this.endpoint = endpoint;
 		this.tenantId = tenantId;
+		d2lfetch.use({ name: 'auth', fn: auth });
```
rcausey, line 13: Can you try creating a separate instance to use in this class? For instance:

```
this.d2lfetch = d2lfetch.addTemp({ name: 'auth', fn: auth });
...
await this.d2lfetch.fetch(request);
...
```

I'm worried about setting it on the global `window.d2lfetch` object, which could interfere/cause issues when it's used in another app that uses d2lfetch.

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -29,14 +33,22 @@ export class Uploader {
 		}
 	}
 
+	defaultLoadingBar(progress) {
+		this.uploadProgress =  progress / (this.waitForProcessing ? 2 : 1);
```
rcausey, line 37: Nit: Looks like there's 2 spaces after `=`. Can you look into whether we can add a linting rule to catch this automatically?

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -96,7 +105,7 @@ export class Uploader {
 						contentDisposition: 'auto'
 					}),
 				onProgress: progress => {
-					this.uploadProgress = progress / (this.waitForProcessing ? 2 : 1);
+					loadingBar.call(this, progress);
```
rcausey, line 108: Does it work if we call it like `this.loadingBar(progress)`? I'd prefer that syntax unless there's a reason to do it this way.

core/d2l-content-uploader/src/util/media-type-util.js
```diff
@@ -2,7 +2,9 @@ const SUPPORTED_AUDIO_EXTENSIONS = ['.m4a', '.mp3', '.ogg', '.wav', '.wma'];
 const SUPPORTED_VIDEO_EXTENSIONS = ['.avi', '.f4v', '.flv', '.m4v', '.mov', '.mp4', '.webm', '.wmv'];
 
 export function getExtension(filePath) {
-	return filePath.split('.').pop().toLowerCase();
+	const pathParts = filePath.split('.');
+	// ensure that does not mistake an extensionless name for a type
+	return pathParts.length > 1 ? pathParts.pop().toLowerCase() : 'missingExtension';
```
rcausey, line 7: Could we throw (and catch) an error instead of returning 'missingExtension'?

rcausey: Or maybe better - just return `''` and handle that case where this is called?

MichaelKovalchuk: Can it return just `null` - no extension?

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -15,10 +15,14 @@ export class Uploader {
 
 		this.uploadProgress = 0;
 
-		this.uploadFile = flow((function * (file, title) {
-			/* eslint-disable no-invalid-this */
-			yield this._uploadWorkflowAsync(file, title);
+		/* eslint-disable no-unused-vars */
+		/* eslint-disable no-invalid-this */
+		// type is unused here, but some references pass it in
+		this.uploadFile = flow((function * (file, title, type = '', loadingBar = this.defaultLoadingBar, loadingBarProcessingWait  = this.defaultLoadingBarProcessingWait) {
```
rcausey, line 21: I might not be fully understanding how this is used (let's chat Monday if not), but could we emit an event when progress is updated instead of overriding these functions?

core/d2l-content-uploader/src/state/uploader.js
willisching: I tried `this.loadingBar(progress)`, but unfortunately it doesn't work

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -15,10 +15,14 @@ export class Uploader {
 
 		this.uploadProgress = 0;
 
-		this.uploadFile = flow((function * (file, title) {
-			/* eslint-disable no-invalid-this */
-			yield this._uploadWorkflowAsync(file, title);
+		/* eslint-disable no-unused-vars */
+		/* eslint-disable no-invalid-this */
+		// type is unused here, but some references pass it in
+		this.uploadFile = flow((function * (file, title, type = '', loadingBarFunction = this.defaultLoadingBarFunction) {
```
rcausey, line 21: Can we change the name `loadingBarFunction` to `progressHandler`? And instead of having a default handler, could we always set `this.uploadProgress` in the object when progress is updated, whether or not you add a special `progressHandler`?

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -15,11 +15,14 @@ export class Uploader {
 
 		this.uploadProgress = 0;
 
+		/* eslint-disable no-invalid-this */
 		this.uploadFile = flow((function * (file, title) {
-			/* eslint-disable no-invalid-this */
-			yield this._uploadWorkflowAsync(file, title);
-			/* eslint-enable no-invalid-this */
+			yield this._ulpoadWorkflowAsync(file, title);
```
rcausey, line 20: Typo `ulpoad`

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -15,11 +15,14 @@ export class Uploader {
...
+		/* eslint-enable no-invalid-this */
```
rcausey, line 21: Is this in the right place?

rcausey, line 21: Is this in the right place?

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -15,11 +15,14 @@ export class Uploader {
...
 		}));
+
+		// allow a way to set special progress handlers such as for smart-curriculum
+		this.progressHandler = function() {};
```
rcausey, line 25: I think `progress` should be an argument to the function right?

rcausey, line 25: I think `progress` should be an argument to the function right?

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -45,11 +48,10 @@ export class Uploader {
 		try {
 			const progress = await this.apiClient.getWorkflowProgress({
 				contentId: this.content.id,
-				revisionId: this.revision.id
+				revisionId: this.revision.id,
 			});
-
 			this.uploadProgress = 50 + ((progress.percentComplete || 0) / 2);
-
+			this.progressHandler();
```
rcausey, line 54: As above, this should be called with `this.uploadProgress` right?

rcausey, line 54: As above, this should be called with `this.uploadProgress` right?

rcausey (dismissed):
Looks good, just a couple small things

core/d2l-content-uploader/src/state/uploader.js
rcausey: Actually, how are you planning on overriding the progressHandler in smart-curriculum? I thought it would still be an argument here?

core/d2l-content-uploader/src/state/uploader.js
rcausey: Related to comment above, why not pass this in as an optional argument like it was before?

rcausey:
Thursday is dev done, so maybe we should wait until after branching to merge this in case it introduces any upload bugs

core/d2l-content-uploader/src/state/uploader.js
willisching: progressHandler is now being set in smart-curriculum with 
`		this.uploader.progressHandler = function () {
			Dispatcher.dispatch({
				actionType: ActionNames.FILE_UPLOAD_PROGRESS,
				bytesUploaded: this.uploadProgress,
				bytesTotal: 100,
			}, /* prevent global refresh: */ true );
		};`
before calling `uploadFile`

rcausey: I think my previous comment might have been kind of confusing. Let me know if you want to discuss. I'm probably taking tomorrow off to watch my daughter, but should be able to jump on a quick call when she's napping.

core/d2l-content-uploader/src/state/uploader.js
```diff
@@ -6,25 +6,22 @@ import { getExtension } from '../util/media-type-util.js';
...
 export class Uploader {
-	constructor({ apiClient, onSuccess, onError, waitForProcessing }) {
+	constructor({ apiClient, onSuccess, onError, waitForProcessing, onProgress = (progress) => {}}) {
+		/* eslint-enable no-unused-vars */
 		this.apiClient = apiClient;
 		this.onSuccess = onSuccess;
 		this.onError = onError;
 		this.waitForProcessing = waitForProcessing;
-
+		this.progressHandler = onProgress;
```
rcausey, line 17: Nit: let's name it `this.onProgress` for consistency.

rcausey (approved):
Looks good, just check my comment

PR #299, opened 2022-04-20
Add audio support for Media Library

clemencetbk:


applications/d2l-capture-central/src/mixins/content-search-mixin.js
```diff
@@ -61,7 +61,7 @@ export const contentSearchMixin = superClass => class extends superClass {
 			contentType: 'video',
 			clientApps: clientApps.join(','),
 			createdAt: createdAt,
-			includeThumbnails: true,
+			includeThumbnails: false,
```
MarquisLP, line 64: Is there a reason we're turning thumbnails off for deleted content?

MarquisLP (commented):
Looks good, I just had a few questions.

applications/d2l-capture-central/src/mixins/content-search-mixin.js
clemencetbk: I think I made it like that initially to convey that the content is deleted/not viewable. But we could turn it back on, would be easier to recognize the videos

applications/d2l-capture-central/locales/en.js
```diff
@@ -57,7 +57,7 @@ export const val = {
 	editProperties: 'Edit Properties',
 	enableChat: 'Enable Chat',
 	enterAnItemName: 'Enter an item name',
-	everyonesVideos: 'Everyone\'s Videos',
+	everyonesMedia: 'Everyone\'s Audio and Video',
```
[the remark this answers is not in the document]
MarquisLP: Oh, I see. Might be good to confirm with Michael and Peter, as I think Peter said yesterday that Media would be better.

applications/d2l-capture-central/src/mixins/content-search-mixin.js
MarquisLP: I see, this does make it clearer that the videos are deleted. I can see benefits for both approaches, though. @MichaelKovalchuk what do you think about showing thumbnails for deleted videos?

MichaelKovalchuk: If we can show thumbnails - lets show them. It makes it easier for users to recognize the content that they may want to restore or delete forever

PR #305, opened 2022-04-27
feat: switching npm registry from artifactory to code artifact

will-malisch:
modified:   .github/workflows/ci.yml


devpow112:
You'll need to switch the component name to `@d2l/content-components` for it to work with code artifact as well. The `@d2l` score is what it is assuming the package will be under.
https://github.com/Brightspace/content-components/blob/master/package.json#L2

PR #308, opened 2022-05-02
Change to support new version of http client

rcausey:
I made a change to the browser http client to not add a `?` at the end of the request URL if the query is empty, so I'm fixing the URL in the d2l-content-properties demo to reflect that change.

I also noticed title/description could be added to the response body of the demo getItem request, so I changed that as well.


core/d2l-content-properties/es-dev-server.config.cjs
```diff
@@ -1,6 +1,8 @@
 const contentServiceEndpointMap = {
-	'/contentservice/api/97dc5895-cd3b-465f-8d5b-3c8d7e1bb6a7/content/a0605fa8-a348-4361-83b0-925fc57e761e?': {
+	'/contentservice/api/97dc5895-cd3b-465f-8d5b-3c8d7e1bb6a7/content/a0605fa8-a348-4361-83b0-925fc57e761e': {
```
rcausey, line 2: New client no longer adds `?` when there's no querystring.

lchan131:
Looks good to me.

lchan131 (approved):

PR #311, opened 2022-05-06
Show only Video and Audio contentTypes

gordon-plumb:
Removing `M4a,Mp3,M4v,Mp4,Mov` from the search filter.


applications/d2l-capture-central/src/util/content-service-client.js
```diff
@@ -184,7 +183,7 @@ export default class ContentServiceClient {
 		size = 15,
 		sort = 'updatedAt:desc',
 		query = '',
-		contentType = [],
+		contentType = '',
 		clientApps = 'all',
 		updatedAt = '',
 		createdAt = '',
```
gordon-plumb, line 189: Reverting this to a string to make it consistent with the other params now that we no longer have to map over the content types.

MichaelKovalchuk (approved):

PR #317, opened 2022-05-12
Show more informative langterm on preview page after uploading Video/Audio during processing

rcausey:
Matches old component behaviour

### DIFF
--- a/core/d2l-content-media-player/d2l-content-media-player.js
+++ b/core/d2l-content-media-player/d2l-content-media-player.js
@@ -33,7 +33,8 @@ class ContentMediaPlayer extends InternalLocalizeMixin(LitElement) {
 			contextId: { type: String, attribute: 'context-id' },
 			contextType: { type: String, attribute: 'context-type' },
 			d2lrn: { type: String, attribute: 'd2lrn' },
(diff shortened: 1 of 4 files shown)


core/d2l-content-media-player/d2l-content-media-player.js
```diff
@@ -33,7 +33,8 @@ class ContentMediaPlayer extends InternalLocalizeMixin(LitElement) {
 			contextId: { type: String, attribute: 'context-id' },
 			contextType: { type: String, attribute: 'context-type' },
 			d2lrn: { type: String, attribute: 'd2lrn' },
-			framed: { type: Boolean, value: false, attribute: 'framed' }
+			framed: { type: Boolean, value: false, attribute: 'framed' },
+			inserting: { type: Boolean }
```
rcausey, line 37: Don't love this parameter name but can't think of anything better. It's just meant to indicate we're in an insertion workflow and should show a more informative processing message (and maybe other stuff in the future)

qwangd2l: An alternative I would suggest is a string variable for a state enum?

PR #320, opened 2022-05-18
content selector component

lchan131:


rcausey:
This looks good, but I'm thinking it might be better if we extracted the buttons out of the individual components and implemented them inside of `d2l-content-selector` instead. This would make the individual components more flexible and re-useable outside of a dialog. Would this be a lot of work to change?

lchan131:
> This looks good, but I'm thinking it might be better if we extracted the buttons out of the individual components and implemented them inside of `d2l-content-selector` instead. This would make the individual components more flexible and re-useable outside of a dialog. Would this be a lot of work to change?

No this would be easy to do. I'll work on it

core/d2l-content-selector/d2l-content-selector.js
```diff
@@ -1,9 +1,276 @@
...
+	LIST: 'list',
+	SETTINGS: 'settings',
+	UPLOAD: 'upload',
+	PROPERTIES: 'properties'
+});
+
+class ContentSelector extends InternalLocalizeMixin(LitElement) {
+	static get properties() {
+		return {
+			allowExistingObjectSelection: { type: Boolean },
```
rcausey, line 20: Same as other comments - I think we should remove this `allowExistingObjectSelection` for now and just default to the behaviour in the SCORM dialog

core/d2l-content-selector/d2l-content-selector.js
```diff
@@ -1,9 +1,276 @@
...
+
+	handleListUpload() {
+		this._selectedView = VIEW.UPLOAD;
+	}
+
+	async handlePropertiesSaved() {
+		this._saveButtonPropertiesDisabled = true;
+		await this.contentProperties.save();
+		this._saveButtonPropertiesDisabled = false;
+		if (this.autoSelectAfterUpload) {
```
rcausey, line 221: Should be removed

core/d2l-content-selector/d2l-content-selector.js
```diff
@@ -1,9 +1,276 @@
...
+			return;
+		}
+
+		this._selectedView = VIEW.LIST;
+	}
+
+	async handleSettingsAddContent() {
+		const content = this.topicSettings.getContent();
+		await this.client.topic.postItem(content);
+		this._openPreview();
```
rcausey, line 232: I don't think openPreview should be done here anymore

core/d2l-content-selector/d2l-content-selector.js
```diff
@@ -1,9 +1,276 @@
...
+	}
+
+	handleSettingsBack() {
+		this._selectedView = VIEW.LIST;
+	}
+
+	handleUploadSuccess({ detail: { d2lrn } }) {
+		if (this.editPropertiesAfterUpload) {
```
rcausey, line 240: Default to SCORM dialog behaviour here instead of using these options

core/d2l-content-topic-settings/d2l-content-topic-settings.js
```diff
@@ -111,6 +120,9 @@ class ContentTopicSettings extends InternalLocalizeMixin(LitElement) {
 
 		this.d2lrn = '';
 		this.serviceUrl = '';
+		this.contentId = '';
+		this.tenantId = '';
+		this.resourceType = '';
```
rcausey, line 125: See other comments in other component about changing this from resourceType to revisionTag and determining resourceType from the revision

rcausey (changes requested):
In general, this looks really good. Most of my comments are just because I've changed my mind from the initial spec and want to keep this as simple as possible until we need those other options.

core/d2l-content-selector/d2l-content-selector.js
lchan131: The _openPreview function only emits an event, so I think it should stay.

core/d2l-content-topic-settings/d2l-content-topic-settings.js
```diff
@@ -92,8 +94,8 @@ class ContentTopicSettings extends InternalLocalizeMixin(LitElement) {
 			line-height: 1.2rem;
 		}
 
-		.action-group {
-			margin-top: 10px;
+		.wrapper {
```
rcausey, line 97: I think this wrapper shouldn't be necessary in here since we moved the buttons outside of the component, right?

core/d2l-content-topic-settings/d2l-content-topic-settings.js
```diff
@@ -122,24 +128,35 @@ class ContentTopicSettings extends InternalLocalizeMixin(LitElement) {
...
 		this.content = content;
 
+		let revision;
+		if (this.revisionTag === 'latest') {
+			revision = content.revisions[content.revisions.length - 1];
+		} else {
+			revision = content.revisions.find(rev => rev.id === this.revisionTag);
+		}
+
+		console.log(revision);
```
rcausey, line 150: Console log

core/d2l-content-topic-settings/d2l-content-topic-settings.js
```diff
@@ -122,24 +128,35 @@ class ContentTopicSettings extends InternalLocalizeMixin(LitElement) {
...
+
+		this._resourceType = revision.type.toLowerCase();
 		this._title = content.title;
 		this._isLoading = false;
 	}
 
 	render() {
 		return html`
-			<div>
+			<div class="wrapper">
```
rcausey, line 159: See comment above re: wrapper

core/d2l-content-properties/demo/index.html
```diff
@@ -17,7 +17,7 @@
 			<h2>d2l-content-properties</h2>
 			<d2l-demo-snippet>
 				<d2l-content-properties
-					d2lrn="d2l:brightspace:content:region:97dc5895-cd3b-465f-8d5b-3c8d7e1bb6a7:scorm:a0605fa8-a348-4361-83b0-925fc57e761e"
+					d2lrn="d2l:brightspace:content:region:97dc5895-cd3b-465f-8d5b-3c8d7e1bb6a7:scorm:0"
```
rcausey, line 20: Doesn't really matter, but just curious why you changed the contentId from a GUID to a number?

core/d2l-content-properties/d2l-content-properties.js
```diff
@@ -98,11 +102,21 @@ class ContentProperties extends InternalLocalizeMixin(LitElement) {
 			margin-top: 10px;
 			font-size: 16px;
 		}
+
+		.settings-container {
```
rcausey, line 106: Let's move this margin out of the component and into the selector instead

core/d2l-content-selector/d2l-content-selector.js
```diff
@@ -1,9 +1,259 @@
...
+		#uploader-wrapper {
+			padding-bottom: 50px;
+		}
+		`;
+	}
+
+	constructor() {
+		super();
+
+		this.allowExistingObjectSelection = true;
```
rcausey, line 58: Let's get rid of `allowExistingObjectSelection` for now

core/d2l-content-selector/d2l-content-selector.js
```diff
@@ -1,9 +1,259 @@
...
+
+		this._nextButtonSettingsDisabled = true;
+		this._saveButtonPropertiesDisabled = false;
+		this._selectedView = VIEW.LIST;
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+
+		if (this.allowUpload && !this.allowExistingObjectSelection) {
```
rcausey, line 81: See above comment on `allowExistingObjectSelection` - shouldn't need this after removal

rcausey (approved):
Looks good, just look into my comments before merging

core/d2l-content-topic-settings/d2l-content-topic-settings.js
lchan131: Again, this is necessary for the demo

core/d2l-content-properties/demo/index.html
lchan131: I just figured it would be easier in the future if we wanted to add a new type, for instance.

core/d2l-content-properties/d2l-content-properties.js
lchan131: Alright. I'll do this for the wrapper classes discussed above as well.

PR #316, opened 2022-05-11
d2l content scorm player

lchan131:

PR #321, opened 2022-05-18
Revert "d2l content scorm player"

rcausey:
Reverts Brightspace/content-components#316


rcausey:
Just temporary so I can make a late add

MichaelKovalchuk (approved):

PR #327, opened 2022-05-20
Use shared clients instead of custom ones in Producer

rcausey:


rcausey:
Depends on https://github.com/Brightspace/content-service-libraries/pull/40

MichaelKovalchuk (approved):

PR #335, opened 2022-05-31
content properties tests

lchan131:


build/iot-client.js
```diff
@@ -39,7 +39,7 @@ var global$1 = (typeof global !== "undefined" ? global :
   typeof self !== "undefined" ? self :
   typeof window !== "undefined" ? window : {});
 
-var buffer = {};
+var buffer$1 = {};
```
rcausey, line 42: Any idea why this file is getting modified?

core/d2l-content-properties/test/d2l-content-properties.test.js
```diff
@@ -0,0 +1,55 @@
...
+		it('should construct', () => {
+			runConstructor('d2l-content-properties');
+		});
+	});
+
+	describe('d2l-content-properties', () => {
+		it('should call save', async() => {
+			const saveStub = Sinon.stub(el, 'save');
+			el.save();
+			assert.isTrue(saveStub.called);
```
rcausey, line 52: Do we need `saveStub` since we're stubbing ContentApi's `getItem` method? Could we just verify that the `getItem` stub was called? And also add a check to make sure `getItem` is called with the parameters we expect (id, revisionTag)

rcausey (changes requested):
Just a couple small things

build/iot-client.js
lchan131: This modification was caused by the visual diff CI creating a new PR. After merging that PR, this change appeared.

core/d2l-content-properties/test/d2l-content-properties.test.js
lchan131: The save method doesn't call `getItem`, but rather `updateItem`. I guess I'll stub that instead of the save method.

rcausey: Sounds good. Let's verify getItem is called with the correct parameters when the component is loaded as well.

build/iot-client.js
rcausey: Hmm.. let me look around to see why this might be happening

rcausey: Can you reset the changes to this file so it's not included in your PR? We're actually going to be moving this out of the repo soon.

core/d2l-content-properties/test/d2l-content-properties.test.js
```diff
@@ -0,0 +1,55 @@
...
+		it('should construct', () => {
+			runConstructor('d2l-content-properties');
+		});
+	});
+
+	describe('d2l-content-properties', () => {
+		it('should call save', async() => {
+			const updateItemStub = Sinon.stub(ContentApi.prototype, 'updateItem');
+			el.save();
+			assert.isTrue(updateItemStub.called);
```
rcausey, line 52: So this verifies that updateItem is called, but it doesn't verify that it's called with the parameters we expect. You can either add the parameters to the stub (`stub.withArgs(...)`), or verify them from the stub afterwards. The same verification should be done with the `getItem` call.

rcausey: Another test to add would be to modify some field input (e.g. package name) and make sure that when save() is called, it gets called with the updated parameter.

lchan131: Done

rcausey: Looks good, but I think you missed my second comment above^

PR #337, opened 2022-06-01
Bump content-service-libraries + fixes

clemencetbk:


applications/d2l-capture-central/src/components/videos/content-list.js
```diff
@@ -267,7 +267,7 @@ class ContentList extends CaptureCentralList {
 			for (let i = 0; i < this._videos.length; i++) {
 				if (this._videos[i].id === contentId) {
 					this._videos[i].processingStatus = 'ready';
-					this._videos[i].poster = await this.apiClient.getPoster({contentId, revisionId});
+					this._videos[i].poster = await this.apiClient.content.getResource({contentId, revisionId});
```
rcausey, line 270: Thanks for catching this. `getResource` is generic to support any type of resource, and the parameters for ID/revision are a bit different, so this needs to change a bit. I think this is correct:

```
this.apiClient.content.getResource({
  id: contentId,
  revisionTag: revisionId,
  resource: 'poster',
  outputFormat: 'signed-url'
});
```

MichaelKovalchuk (approved):

applications/d2l-capture-central/src/components/videos/content-list.js
```diff
@@ -276,10 +281,10 @@ class ContentList extends CaptureCentralList {
 	}
 
 	async _setUpIotClient() {
-		const { accessKeyId, secretAccessKey, sessionToken, region } = await this.apiClient.requestNotificationsClientAuth();
-		const host = (await this.apiClient.getIotEndpoint()).iotEndpoint;
+		const { accessKeyId, secretAccessKey, sessionToken, region } = await this.apiClient.notifications.requestNotificationsClientAuth();
+		const host = (await this.apiClient.notifications.getIotEndpoint()).iotEndpoint;
 		new IotClient({
-			tenantId: this.tenantId,
+			tenantId: this.apiClient.tenant.tenantId,
```
rcausey, line 287: Can we get the tenantId from `this.requestDependency('tenant-id');` instead?

MichaelKovalchuk, line 287: we are referencing tenantId like this at the line 36: https://github.com/Brightspace/content-components/blob/70cdf0feebad15737872a324d28bfee3e9b10121/applications/d2l-capture-central/src/components/videos/content-list.js#L36
`this.tenantId = this.apiClient.tenantId;`
but here we use `this.apiClient.tenant.tenantId`
why it we could not use this.tenantId?

rcausey: See my previous message. We should use `this.tenantId` but get it from `this.requestDependency('content-service-client')` instead of from the client, since it's not really meant to be a public property of the client and might change in the future.

rcausey (approved):

PR #352, opened 2022-06-06
feat: d2l-content-topic-settings tests

lchan131:


core/d2l-content-topic-settings/d2l-content-topic-settings.js
```diff
@@ -162,27 +162,27 @@ class ContentTopicSettings extends InternalLocalizeMixin(LitElement) {
 						</h4>
 						<div class="setting-option">
 							<input
-								id="latest-version"
+								id="create-grade-item-yes"
```
rcausey, line 165: Can you dispatch a "change" event whenever any of the inputs change in the d2l-content-topic-settings component which includes all of the topic settings data that we need to make the `topic.postItem` call outside of the component?

And also write some tests to verify the event is dispatched when/as expected?

rcausey (changes requested):

core/d2l-content-topic-settings/d2l-content-topic-settings.js
lchan131: As discussed in Slack, I did not emit an event, but made sure that the getContent method is working as expected when inputs are changed.

rcausey (approved):
Looks good, but please revert build/iot-client.js before merging

PR #354, opened 2022-06-06
feat: Version bump for thumbnail solution

will-malisch:


MichaelKovalchuk:
this change is already in 142 version of BSI. Just verified

